PR #6, opened 2015-10-22
Authenticated user gets kicked when trying to connect with the same account with wrong credentials

galvezlj:
Authenticated user gets kicked when trying to connect another client instance with the same account even with wrong password.

### DIFF
--- a/lib/handlers/connect.js
+++ b/lib/handlers/connect.js
@@ -15,6 +15,7 @@ function ClientPacketStatus (client, packet) {
 
 var connectActions = [
   authenticate,
+  registerClient,
   fetchSubs,
   restoreSubs,
   storeWill,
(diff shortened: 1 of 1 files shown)


lib/handlers/connect.js
```diff
@@ -34,7 +35,7 @@ function handleConnect (client, packet, done) {
   client.id = packet.clientId || uuid.v4()
   client.will = packet.will
 
-  client.broker.registerClient(client)
+  //client.broker.registerClient(client)
```
mcollina: can you please remove this commented line?

mcollina:
Can you please add a unit test for this?
Also, the CI build is failing: https://travis-ci.org/mcollina/aedes/jobs/86800040

galvezlj:
This is incomplete as it prevents users to subscribe properly after authentication. Need to check deeper into the authentication procedure to check what it needs from the registration procedure.

mcollina:
Thanks, would you mind adding a couple of unit tests as well?

PR #11, opened 2015-11-21
Update authorizeSubscribe documentation in README

platy:
The example suggested that the authorizeSubscribe method receives the topic as a parameter (like in mosca) but it takes a subscription object.


platy:
I've only changed the README so I don't know why the build failed

mcollina:
Thanks!! :)

mcollina:
I updated fastparallel, and there was a bug in a test.

PR #13, opened 2015-12-05
Fix require() statement in README example

eladnava:
The README.md example code does not run as-is unless the `require('./aedes')` is changed to `require('aedes')` since the module is installed via `npm install aedes --save`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ npm install aedes --save
 <a name="example"></a>
 ## Example
 ```js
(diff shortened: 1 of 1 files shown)


mcollina:
Thanks for the good catch!!!

eladnava:
@mcollina My pleasure! Gave `aedes` a try, definitely liking it so far! Might deploy it to production as a replacement to `mosca` soon.

PR #22, opened 2016-01-01
Stop ingesting packets after the client has errored

### LINKED ISSUE #21: Issue remains with edge cases
I still seem to be hitting the same scenario where ```client._keepaliveTimer === null``` at line 19 in index.js

    client._keepaliveTimer.reschedule(client._keepaliveInterval)

It's hard to recreate the exact conditions as this is in our production system, but the packet which is being handled at that moment is...

{"cmd":"puback","retain":false,"qos":0,"dup":false,"length":2,"topic":null,"payload":null,"messageId":30407}

...in case that helps to make sense of it. Otherwise I have to do some fairly heavy lifting to figure out the full sequence of events leading up to this failure.

Here's the stack trace...

```

TypeError: Cannot read property 'reschedule' of null
    at handle (/Users/cefn/Documents/code/imagination/git/aedes/lib/handlers/index.js:19:27)
    at Parser.enqueue (/Users/cefn/Documents/code/imagination/git/aedes/lib/client.js:204:3)
    at Parser._newPacket (/Users/cefn/Documents/code/imagination/git/aedes/node_modules/mqtt-packet/parser.js:30:10)
    at Parser.parse (/Users/cefn/Documents/code/imagination/git/aedes/node_modules/mqtt-packet/parser.js:42:48)
    at DestroyableTransform.nextBatch (/Users/cefn/Documents/code/imagination/git/aedes/lib/client.js:55:23)
    at emitReadable_ (node_modules/websocket-stream/node_modules/readable-stream/lib/_stream_readable.js:448:10)
    at emitReadable (node_modules/websocket-stream/node_modules/readable-stream/lib/_stream_readable.js:444:5)
    at readableAddChunk (node_modules/websocket-stream/node_modules/readable-stream/lib/_stream_readable.js:187:9)
    at DestroyableTransform.Readable.push (node_modules/websocket-stream/node_modules/readable-stream/lib/_stream_readable.js:149:10)
    at DestroyableTransform.Transform.push (node_modules/websocket-stream/node_modules/readable-stream/lib/_stream_transform.js:145:32)
    at WebSocket.onmessage (node_modules/websocket-stream/stream.js:74:11)
    at WebSocket.onMessage (node_modules/websocket-stream/node_modules/ws/lib/WebSocket.js:414:14)
    at Receiver.onbinary (node_modules/websocket-stream/node_modules/ws/lib/WebSocket.js:804:10)
    at node_modules/websocket-stream/node_modules/ws/lib/Receiver.js:533:18
    at node_modules/websocket-stream/node_modules/ws/lib/Receiver.js:357:7
    at node_modules/websocket-stream/node_modules/ws/lib/PerMessageDeflate.js:217:5
    at afterWrite (_stream_writable.js:346:3)
    at onwrite (_stream_writable.js:337:7)
    at WritableState.onwrite (_stream_writable.js:89:5)
    at afterTransform (_stream_transform.js:79:5)
    at TransformState.afterTransform (_stream_transform.js:54:12)
    at Zlib.callback (zlib.js:614:5)
```

As a reference, the same production scenario completes with Mosquitto without errors and we're only engaging with Aedes as a client through MQTT.js (with no deeper integration) so I don't think we're doing anything unusual from our side.

mcollina:
Fixed #21.

@cefn please test and see if it solves your issue.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -158,6 +158,9 @@ Client.prototype.close = function (done) {
     finish()
   }
 
+  this.parser.removeAllListeners('packet')
+  conn.removeAllListeners('readable')
+
   if (this._keepaliveTimer) {
     this._keepaliveTimer.clear()
     this._keepaliveTimer = null
--- a/lib/client.js
+++ b/lib/client.js
@@ -163,6 +163,7 @@ Client.prototype.close = function (done) {
 
   if (this._keepaliveTimer) {
     this._keepaliveTimer.clear()
+    this._keepaliveInterval = -1
     this._keepaliveTimer = null
   }
 
--- a/test/regr-21.js
+++ b/test/regr-21.js
@@ -1,0 +1,30 @@
+'use strict'
+
+var test = require('tape').test
+var helper = require('./helper')
+var setup = helper.setup
+var connect = helper.connect
+
+test('after an error, outstanding packets are discarded', function (t) {
+  t.plan(1)
+  var s = connect(setup(), {
+    keepalive: 1000
+  })
+  var packet = {
+    cmd: 'publish',
+    topic: 'hello',
+    payload: 'world'
+  }
+
+  s.broker.mq.on('hello', function (msg, cb) {
+    t.pass('first msg received')
+    s.inStream.emit('error', new Error('something went wrong'))
+    setImmediate(cb)
+  })
+
+  s.inStream.write(packet)
+  setImmediate(function () {
+    s.inStream.write(packet)
+    s.inStream.write(packet)
+  })
+})


cefn:
Indeed this branch helps our production test suite to proceed beyond the error I reported.

We're now having much subtler issues when running with Aedes (versus Mosquitto) which only appear when interactively testing. I'm investigating if it's to do with failing to honour the Last Will and Testament - what's the status of LWT for Aedes?

It seems that the Last Will and Testament is being written to the topic tree before the normal trigger event (when the browser is shut down) or potentially something is forcing the client to be disconnected and hence correctly triggering the LWT (which is somehow swallowed as an error in our production code).

On the plus side I can see that the topic tree reported by ```mosquitto_sub -t '#' -v``` from a Aedes production test run and a Mosquitto production test run is so similar that I'm having to investigate in some detail what is the difference which makes one work and the other not, which is a pretty good sign. Given the complexity of the production code, there's still the chance it is some kind of race condition given the timings of the brokers responding could be different. Will let you know if I can recreate in a simpler test suite.

cefn:
Forgot to mention for reference that Mosca does satisfy our integration testing suite, but just super-slow owing to performance issues as detailed at https://github.com/mcollina/mosca/issues/385 which which we couldn't easily hide from users.

Aedes is much, much more performant in our integration testing. It's not as fast as the newly-patched version of Mosquitto (Ral helped me eliminate the 200ms round-trip delay, at the cost of burning some cycles in poll loops), but is fast enough in end-to-end behaviour that we could certainly use it in production and has much better potential for provisioning and administering access-control lists from within our node server logic, so we could take the hit on raw performance in favour of these other aspects I think.

For those reasons I'm very interested to keep sweating Aedes until it satisfies our integration testing suite, assuming I can figure out the issue we're hitting right now.

mcollina:
@cefn are you running with some form of keepalive? Might be some form of https://github.com/maxogden/websocket-stream/issues/69.

BTW, I would be really happy if you can shelve off some ms from here as well.

cefn:
I assume that given our production code is Node and V8 in Chrome, running MQTT.js then we should be golden (https://www.npmjs.com/package/mqtt#client and https://github.com/maxogden/websocket-stream/issues/69#issuecomment-122546881 suggest keepalive would be 10 seconds by default in MQTT.js without any special configuration I think).

Generally I have had trouble getting meaningful performance profiling out of V8 so far (so much indirection in the libraries we were using before) but may be better now we've removed almost all our promise and stream code in the latest iteration. With a bit of luck I'll be able to pull out the Aedes functions which run the longest and explore any marginal improvements. Our tests are frequently disadvantaged by a cold-start without optimised hot paths so I'm not sure there's any real performance problem yet.

PR #30, opened 2016-02-08
fixed unsubscribe handler exception

dpatekar:
Unsubscribe request for a topic that hasn't been subscribed is not handeld.


mcollina:
Great! Would you mind to add a little unit test?

mcollina:
Thanks so much!!

PR #32, opened 2016-02-17
prevent null client unregister

dpatekar:
When using TLS like this:
```
var tlsServer = tls.createServer(options, function(s){  
  console.log('secure connection established');
  aedes.handle(s);
});
tlsServer.listen(config.ports.tcptls);
```
Sometimes happens that client does not connect completely (e.g. using some ssl scan tool like https://github.com/rbsec/sslscan) and **end-of-stream** is firing a callback which is **client.close()**. In that situation, client is not connected to the broker. That causes incorrect decrement of connected clients in broker.unregisterClient()

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -75,9 +75,14 @@ function Client (broker, conn) {
   }
 
   this.deliverQoS = function deliverQoS (_packet, cb) {
-    var packet = new QoSPacket(_packet, that)
-    packet.writeCallback = cb
-    broker.persistence.outgoingUpdate(that, packet, writeQoS)
+    //TODO: downgrade to QoS1 if applicable
+    if (_packet.qos == 0){
+      that.deliver0(_packet, cb);
+    } else {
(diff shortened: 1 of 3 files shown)


mcollina:
Can you please add a unit test for this?
Il giorno mer 17 feb 2016 alle 19:56 dpatekar <notifications@github.com> ha
scritto:

dpatekar:
I have added a unit test in the existing one "authenticate unsuccessfully a client with username and password" because that situation reproduces the issue. Just check for connectedClients count on end-of-stream.

In addition to this, I have added a unit test and solution to the situation when a client subscribes to some topic with QoS 1 and then publishes a packet to the same topic with QoS 0. Persistence module is throwing an error in that situation because there is no downgrade to QoS 0 and according to the MQTT specs it should be because subscribing QoS should define the maximum QoS for that topic. What do you think about this?

mcollina:
That's great and awesome! Thanks so much for spotting the QoS downgrading!

PR #36, opened 2016-02-26
Add options for overriding methods in broker creation

pkinney:
`authenticate`, `authorizePublish`, `authorizeSubscribe`, and `published` methods can be passed in the configuration options when initializing the broker or can be set after the fact as it works now:

```javascript
var aedes = require('aedes')({
  authenticate: function (client, username, password, callback) {
    callback(null, username === 'skroob' && password === '12345')
  }
)};
```

Also refactored the options ab it to use a `defaults` function for clarity.

This addresses https://github.com/mcollina/aedes/issues/9


mcollina:
Thanks, releasing soon!

PR #39, opened 2016-03-16
Add 'delivered' event on PUBACK handler

pkinney:
This addresses https://github.com/mcollina/aedes/issues/10 and emits a `delivered` event when a PUBACK is received.

### DIFF
--- a/lib/handlers/puback.js
+++ b/lib/handlers/puback.js
@@ -3,6 +3,7 @@
 function handlePuback (client, packet, done) {
   var persistence = client.broker.persistence
   persistence.outgoingClearMessageId(client, packet, done)
+  client.broker.emit('delivered', packet, client)
 }
 
 module.exports = handlePuback
--- a/lib/handlers/puback.js
+++ b/lib/handlers/puback.js
@@ -3,7 +3,7 @@
 function handlePuback (client, packet, done) {
   var persistence = client.broker.persistence
   persistence.outgoingClearMessageId(client, packet, done)
-  client.broker.emit('delivered', packet, client)
+  client.broker.emit('deliver', packet, client)
 }
 
 module.exports = handlePuback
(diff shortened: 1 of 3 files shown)


mcollina:
Can you please do the same for PUBCOM and add the doc relative to this in the README?
Also, I will call that `'deliver'` (we are already breaking backward compat with mosca :/).

pkinney:
Right now, the `PUBCOMP` case currently falls through to the `PUBACK` case (https://github.com/mcollina/aedes/blob/master/lib/handlers/index.js#L35-L38), so the change to the `handlePuback` function should handle both cases.  I added a test for PUBCOMP to make sure and for regression tests.

I changed the event to 'deliver' and updated the test.  Working on updates to README now.

pkinney:
README updated.  Let me know if there is anything else to add.

test/client-pub-sub.js
```diff
@@ -73,6 +73,77 @@ test('publish direct to a single client QoS 1', function (t) {
...
+  t.plan(3)
+
+  var broker = aedes()
+  var messageId
+
+  broker.on('client', function (client) {
+    client.publish({
+      topic: 'hello',
+      payload: new Buffer('world'),
+      qos: 1
```
mcollina: This should be qos:2

pkinney: Yep. Corrected.

mcollina:
Release as v0.14.0

lib/handlers/puback.js
```diff
@@ -3,6 +3,7 @@
 function handlePuback (client, packet, done) {
   var persistence = client.broker.persistence
   persistence.outgoingClearMessageId(client, packet, done)
+  client.broker.emit('deliver', packet, client)
```
behrad: I believe `delivered` or `ack` could be a better name, in Mosca we are using `delivered` @mcollina

mcollina: I think 'ack' is the best of all, actually. I'll get that updated asap!
Il giorno dom 3 apr 2016 alle 18:14 Behrad <notifications@github.com> ha
scritto:

PR #44, opened 2016-04-26
Added authorizeForward

stoiKris:
Hello. I've added the method we talked about. However, I'm not very sure about the arguments of the callback. Right now, it only looks at err, the rest being modified on the original object.

I did see a slight drop in perfomance, about 2-5% after adding the method, without supplying anything custom in the config.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -69,32 +69,26 @@ function Client (broker, conn) {
   this._eos = eos(this.conn, this.close.bind(this))
 
   this.deliver0 = function deliverQoS0 (_packet, cb) {
-    that.broker.authorizeForward(that, _packet, function (err) {
-      if (err != null) {
-        // todo add an error message
-        return
-      }
-      var packet = new Packet(_packet, broker)
-      packet.qos = 0
-      write(that, packet, cb)
-    })
+    var packet = new Packet(_packet, broker)
+    packet.qos = 0
+    write(that, packet, cb)
   }
 
   this.deliverQoS = function deliverQoS (_packet, cb) {
-    that.broker.authorizeForward(that, _packet, function (err) {
-      if (err != null) {
-        // todo add an error message
-        return
-      }
-      // downgrade to qos0 if requested by publish
-      if (_packet.qos === 0) {
-        that.deliver0(_packet, cb)
-      } else {
-        var packet = new QoSPacket(_packet, that)
-        packet.writeCallback = cb
-        broker.persistence.outgoingUpdate(that, packet, writeQoS)
-      }
-    })
+    var _authPacket = that.broker.authorizeForward(that, _packet)
+    if (util.isNullOrUndefined(_authPacket)) {
+      // we consider this to be an error, since the packet is undefined, so there's nothing to send
+      that.broker.persistence.outgoingClearMessageId(that, _packet, function () {})
+      return
+    }
+    // downgrade to qos0 if requested by publish
+    if (_authPacket.qos === 0) {
+      that.deliver0(_authPacket, cb)
+    } else {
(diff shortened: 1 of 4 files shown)


stoiKris:
Uhm, seems I didn't cover all my bases in the unit test. I could look into it a bit more if that's a problem

mcollina:
A bad design decision I did in Mosca is making that function async. That
function needs to be synchronous, otherwise performance will be extremely
affected.

Would you mind making it sync?
Il giorno mar 26 apr 2016 alle 20:39 stoiKris <notifications@github.com> ha
scritto:

cristian-stoicescu:
I thought it was strange the performance dropped after I added the callback.
Btw, in your opinion, should the function also return the package and client, or just it's parameters, and only error?
I'm thinking an elegant way would be to return undefined if there was an error else return the client and the packet.
What do you think?

mcollina:
@cristian-stoicescu just return the packet, so the function can overwrite anything.

BTW, `authorizeForward` needs to be called only if it's a publish for QoS 1/2. Moreover you will need to call `persistence.outgoingClearMessageId` to remove the packet if the authorization is denied, if the client is not clean. 

I know it's a lot of work for this feature, thanks for helping out!

lib/client.js
```diff
@@ -75,11 +75,17 @@ function Client (broker, conn) {
   }
 
   this.deliverQoS = function deliverQoS (_packet, cb) {
+    var _authPacket = that.broker.authorizeForward(that, _packet)
+    if (util.isNullOrUndefined(_authPacket)) {
+      // we consider this to be an error, since the packet is undefined, so there's nothing to send
+      that.broker.persistence.outgoingClearMessageId(that, _packet, function () {})
```
mcollina: don't allocate a function here.

lib/client.js
```diff
@@ -75,11 +75,17 @@ function Client (broker, conn) {
   }
 
   this.deliverQoS = function deliverQoS (_packet, cb) {
+    var _authPacket = that.broker.authorizeForward(that, _packet)
+    if (util.isNullOrUndefined(_authPacket)) {
```
mcollina: just check `if (_authPacket)`

lib/client.js
```diff
@@ -75,11 +75,17 @@ function Client (broker, conn) {
   }
 
   this.deliverQoS = function deliverQoS (_packet, cb) {
+    var _authPacket = that.broker.authorizeForward(that, _packet)
```
mcollina: you can just name this `packet`.

stoiKris: The linter tells me packet is already defined:D.
Maybe "_toForward" ?

stoiKris:
Sorry for the tons of commits, github has a habit of hiding "outdated" comments, so I see them too late

lib/client.js
mcollina: `_toForward` is ok for me.

mcollina: you should also check if `_packet.cmd` is  `'publish'` before doing this check.

stoiKris: yes, you're right. I didn't realize deliverQoS would deliver everything. I originally thought it only would deliver publishes by default :)

mcollina:
Actually no, there is no need to check, those methods are only used for
publishes :/. Sorry
Il giorno mer 27 apr 2016 alle 13:32 Coveralls <notifications@github.com>
ha scritto:

mcollina:
You need to do the same for the other deliver method, as one is used for QoS 0 and one for QoS 1/2.

stoiKris:
Didn't you say earlier that the `authorizeForward` needs to be called only if it's a publish for QoS 1/2 ?

I'm confused

PR #57, opened 2016-06-17
clarify delivery concurrency is per client

behrad:


mcollina:
This is not true. Delivery concurrency is for the whole broker, and the limit is on the fan in, not on the fan out.

behrad:
and how this is achieved by `conn.setMaxListeners(opts.concurrency * 2)` ?

behrad:
@mcollina Can you clarify me on this man?

mcollina:
A piece of code is probably worth more :).

`conn.setMaxListeners(opts.concurrency * 2)`

is needed to avoid a bad warning. Nothing more. It's set to twice the max concurrency, just because I had to account for QoS 1/2 messages.

regarding the _actual_ concurrency limit, this is is:
https://github.com/mcollina/mqemitter/blob/master/mqemitter.js#L97-L114

mcollina:
It is also invoked in mqemitter-redis: https://github.com/mcollina/mqemitter-redis/blob/master/mqemitter-redis.js#L33.

behrad:
Not it makes sense, thank you :)

implementation shows the limit is number of `parallel publish matchings` done inside the broker, however each publish may match any number of clients :)

mcollina:
Would you mind sending a PR against the README? It might be helpful to others.

behrad:
> implementation shows the limit is number of parallel publish matchings done inside the broker, however each publish may match any number of clients :)

and this means the max number of messages delivered concurrently may be higher than conncurrency!? in current implementation. Yes?

mcollina:
concurrency is the number of _unique_ messages being delivered, not considering the number of clients we are delivering them to.

behrad:
> concurrency is the number of unique messages being delivered, not considering the number of clients we are delivering them to.

But here https://github.com/mcollina/mqemitter/blob/master/mqemitter.js#L126 current is incremented per _do, despite the actual matches size!!!

mcollina:
there is no check on the matches, the check is on the message that is being published, not on the messages it is being delivered.

behrad:
> the check is on the message that is being published, not on the messages it is being delivered.

Exactly! and hence we need to emphasize the word unique message, and may be change the word delivered to `forwarded` or `published` since delivered increases the chance for ambiguous.

mcollina:
Published is a good term I think. PR?

behrad:
👍 Yup, will create one.

PR #62, opened 2016-08-02
Handling for other return codes

GavinDmello:
This has the extra error codes Aedes doesn't return

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,19 @@ instance.authenticate = function (client, username, password, callback) {
   callback(null, username === 'matteo')
 }
 ```
+Other return codes can passed as follows :-
+
+```js
+instance.authenticate = function (client, username, password, callback) {
+  callback({ returnCode: 1 }, null)
+}
+```
+The return code values and their responses which can be passed are given below :-
+
+1 - Unacceptable protocol version
+2 - Identifier rejected
+3 - Server unavailable
+4 - Bad user name or password
 
 -------------------------------------------------------
 <a name="authorizePublish"></a>
--- a/README.md
+++ b/README.md
@@ -172,10 +172,10 @@ instance.authenticate = function (client, username, password, callback) {
 ```
 The return code values and their responses which can be passed are given below :-
 
-1 - Unacceptable protocol version
-2 - Identifier rejected
-3 - Server unavailable
-4 - Bad user name or password
+1.  Unacceptable protocol version 
+2.  Identifier rejected 
+3.  Server unavailable 
+4.  Bad user name or password
 
 -------------------------------------------------------
 <a name="authorizePublish"></a>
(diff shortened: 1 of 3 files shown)


lib/handlers/connect.js
```diff
@@ -63,10 +63,44 @@ function authenticate (arg, done) {
...
+            break
+
+          case 4 :
+          default :
+            write(client, {
+              cmd: 'connack',
+              returnCode: 4
+            }, client.close.bind(client, done))
+            break
+      }
```
mcollina: how about a check on err.returnCode if it's between 1 and 4, and use 4 instead? We can avoid the switch.

mcollina:
Can you please add unit tests and docs?

GavinDmello:
Cool. I'll make the changes :)

GavinDmello:
Will there be test cases for each return code , or a generic one ? Also do I have to add documentation in the Readme ?

mcollina:
I think you should have test cases for each return code, yes. You can set them up via a for loop.

Yes, documentation goes in the README.

Thanks!

lib/handlers/connect.js
```diff
@@ -63,10 +63,26 @@ function authenticate (arg, done) {
...
-        cmd: 'connack',
-        returnCode: 4
-      }, client.close.bind(client, done))
+      if (err && !err.returnCode) {
+        write(client, {
+          cmd: 'connack',
+          returnCode: 4
+        }, client.close.bind(client, done))
+        return
+      } else {
```
mcollina: You can probably avoid the first if.

GavinDmello: That was for checking if somebody doesn't send the returnCode property. I can merge it with the if below though.

mcollina: that's what I meant :)

README.md
```diff
@@ -163,6 +163,19 @@ instance.authenticate = function (client, username, password, callback) {
   callback(null, username === 'matteo')
 }
 ```
+Other return codes can passed as follows :-
+
+```js
+instance.authenticate = function (client, username, password, callback) {
+  callback({ returnCode: 1 }, null)
+}
+```
```
mcollina: Can you please attach the returnCode to an Error object?

README.md
```diff
@@ -163,6 +163,19 @@ instance.authenticate = function (client, username, password, callback) {
...
+The return code values and their responses which can be passed are given below :-
+
+1 - Unacceptable protocol version
+2 - Identifier rejected
+3 - Server unavailable
+4 - Bad user name or password
```
mcollina: Can you please make this bullet points, and wrap the numbers with backtick?

README.md
GavinDmello: Do you mean assigning it to an error variable ?

mcollina:
That's perfect, thanks very much!

PR #65, opened 2016-09-01
Emit connackSent event

rahulbhanushali:
Emitted a connackSent event to allow implementing operations required after connack has been sent.

### DIFF
--- a/test/basic.js
+++ b/test/basic.js
@@ -37,42 +37,6 @@ test('connect and connack (minimal)', function (t) {
   })
 })
 
-test('connect and connackSent event', function (t) {
-  var s = setup()
-  var clientId = 'my-client'
-
-  t.plan(2)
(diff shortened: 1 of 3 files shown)


mcollina:
This will need a unit test

test/basic.js
```diff
@@ -37,6 +37,42 @@ test('connect and connack (minimal)', function (t) {
...
+      qos: 0,
+      retain: false,
+      dup: false,
+      topic: null,
+      payload: null,
+      sessionPresent: false
+    }, 'successful connack')
+    t.end()
+  })
+})
```
mcollina: Can you please move this at the end of https://github.com/mcollina/aedes/blob/master/test/meta.js?

mcollina:
Thanks!

PR #69, opened 2016-09-30
presence module for aedes

GavinDmello:
An online/offline feature for aedes

### DIFF
--- a/aedes.js
+++ b/aedes.js
@@ -223,6 +223,7 @@ var publishFuncsQoS = [
   callPublished
 ]
 Aedes.prototype.publish = function (packet, client, done) {
+  //console.log(packet, client, done)
   if (typeof client === 'function') {
     done = client
     client = null
@@ -270,6 +271,10 @@ Aedes.prototype.unregisterClient = function (client) {
   this.connectedClients--
   delete this.clients[client.id]
   this.emit('clientDisconnect', client)
+  this.publish({
+    topic: '$SYS/' + this.id + '/new/clientDisconnect',
+    payload: new Buffer(client.id, 'utf8')
+  }, noop)
 }
 
 function closeClient (client, cb) {
--- a/aedes.js
+++ b/aedes.js
@@ -223,7 +223,6 @@ var publishFuncsQoS = [
   callPublished
 ]
 Aedes.prototype.publish = function (packet, client, done) {
-  //console.log(packet, client, done)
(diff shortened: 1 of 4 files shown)


test/qos2.js
```diff
@@ -129,9 +129,10 @@ test('call published method with client with QoS 2', function (t) {
   }
 
   broker.published = function (packet, client, cb) {
-    t.ok(client, 'client must be passed to published method')
-
-    cb()
+    if (packet.topic.split('/').pop() !== 'clientDisconnect') {
```
GavinDmello: This test case was failing because the client object was made null after Aedes publish. Let me know if you have a better way of going about this.

aedes.js
```diff
@@ -223,6 +223,7 @@ var publishFuncsQoS = [
   callPublished
 ]
 Aedes.prototype.publish = function (packet, client, done) {
+  //console.log(packet, client, done)
```
mcollina: Can you please remove this?

aedes.js
```diff
@@ -270,6 +271,10 @@ Aedes.prototype.unregisterClient = function (client) {
   this.connectedClients--
   delete this.clients[client.id]
   this.emit('clientDisconnect', client)
+  this.publish({
+    topic: '$SYS/' + this.id + '/new/clientDisconnect',
+    payload: new Buffer(client.id, 'utf8')
+  }, noop)
```
mcollina: I think this should be named differently. Would you mind checking what's the situation in Mosquitto and Mosca?

test/qos2.js
mcollina: Yes. but we should put a note here, and maybe in the docs too. For server publishes the client is always null.

GavinDmello: I think its already mentioned in the published method block. I'll add it here though.

aedes.js
GavinDmello: Mosquitto doesn't provide this infomation. It provides the number of connected/disconnected clients. Mosca uses $SYS/+/new/clients &  $SYS/+/disconnect/clients  for connections & disconnections.

mcollina: use the same as mosca then :)

GavinDmello:
Would this require a change in the Readme ?

mcollina:
Yes, please.

GavinDmello:
Done.

mcollina:
Good Job.

PR #70, opened 2016-10-03
Docs: instance.publish(packet) expects an aedes-packet object

eladnava:
Updated the docs to match the new parameter expectation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,10 +142,22 @@ It supports backpressure.
 <a name="publish"></a>
 ### instance.publish(packet, done)
 
-Publish the given packet to subscribed clients and functions. A packet
-must be valid for [aedes-packet](http://npm.im/aedes-packet).
+Publish the given packet to subscribed clients and functions. It supports backpressure.
 
-It supports backpressure.
+A packet must contain the following properties:
+
+```js
+{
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -143,7 +143,7 @@ It supports backpressure.
 ### instance.publish(packet, done)
 
 Publish the given packet to subscribed clients and functions. A packet
-must be valid for [mqtt-packet](http://npm.im/mqtt-packet).
+must be valid for [aedes-packet](http://npm.im/aedes-packet).
```
mcollina: In aedes-packet, everything is defaulted:

https://github.com/mcollina/aedes-packet/blob/master/packet.js#L4-L11.

I think a much better definition can be: a packet must have the following properties...

eladnava: @mcollina I think you have outlined it well here for the Publish packet:
https://www.npmjs.com/package/mqtt-packet#publish

Should I just copy over that explanation?

mcollina: I think so, yes!

eladnava: @mcollina Updated the PR =) Let me know if anything is missing.

mcollina:
Perfect, thanks!

PR #75, opened 2016-10-25
Emit an event when the server receives a ping

renatoc:


renatoc:
@mcollina Yes, i will do that.

renatoc:
@mcollina Thanks!

PR #78, opened 2016-12-17
tweak readme

ralphtheninja:


README.md
```diff
@@ -257,7 +257,6 @@ instance.authorizeSubscribe = function (client, sub, cb) {
 
 It will be called when a client is set to recieve a message. Override to supply custom
 authorization logic.
-In order to completely pre
```
ralphtheninja: A leftover sentence? Just removed it since I didn't know how to finish it.

README.md
```diff
@@ -159,7 +159,7 @@ A packet must contain the following properties:
 }
 ```
 
-Only the `topic` and properties are mandatory
+Only the `topic` and `payload` properties are mandatory.
```
ralphtheninja: I guess this one was copy/pasted from `mqtt-packet` repo.

README.md
```diff
@@ -147,7 +147,7 @@ It supports backpressure.
 
 Publish the given packet to subscribed clients and functions. It supports backpressure.
 
-A packet must contain the following properties:
+A packet contains the following properties:
```
ralphtheninja: Contradicting sentences. This one states that all properties are mandatory but the second says only topic (and as I suspect, the payload property) is mandatory.

README.md
ralphtheninja: I guess `payload` is missing?

ralphtheninja:
_bump_

README.md
mcollina: You are right

README.md
mcollina: Good call

PR #82, opened 2017-01-27
$SYS topic to get the number of connected clients.

debjeet-sarkar:


aedes.js
```diff
@@ -257,23 +257,36 @@ Aedes.prototype.registerClient = function (client) {
...
   this.connectedClients++
   this.clients[client.id] = client
   this.emit('client', client)
   this.publish({
     topic: '$SYS/' + this.id + '/new/clients',
     payload: new Buffer(client.id, 'utf8')
   }, noop)
+
+  this.publish({
+  topic: '$SYS/' + this.id + '/clients/total',
```
GavinDmello: I think it should be $SYS/broker/clients/total and it should be a retained message

mcollina:
I do not think this is the correct approach, but rather we should the same pattern we have for Mosca, https://github.com/mcollina/mosca/blob/master/lib/stats.js.

In fact I even created https://github.com/mcollina/aedes-stats, but it is empty :(. If you want to contribute things there, I would be very happy.

debjeet-sarkar:
@mcollina, makes sense.

PR #86, opened 2017-02-06
Added closed event on server close

debjeet-sarkar:


mcollina:
@debjeet-sarkar can you add a unit test? also can you rebase the commit? and add it to the README?

aedes.js
```diff
@@ -283,6 +283,7 @@ function closeClient (client, cb) {
 Aedes.prototype.close = function (cb) {
   clearInterval(this._heartbeatInterval)
   clearInterval(this._clearWillInterval)
+  this.emit('closed')
   this._parallel(this, closeClient, Object.keys(this.clients), cb || noop)
```
mcollina: can you move the `emit` inside the `_parallel` callback.

debjeet-sarkar: Sure

aedes.js
```diff
@@ -283,7 +283,16 @@ function closeClient (client, cb) {
 Aedes.prototype.close = function (cb) {
   clearInterval(this._heartbeatInterval)
   clearInterval(this._clearWillInterval)
-  this._parallel(this, closeClient, Object.keys(this.clients), cb || noop)
+  this._parallel(this, closeClient, Object.keys(this.clients), doneClose(this, cb))
```
mcollina: we should create a closure here, and call the callback when it's called.

PR #89, opened 2017-02-24
max listeners independent of concurrency

GavinDmello:


GavinDmello:
Why doesn't it make sense ? 
What point it does it hang ?
I'm able to publish 20,000+ messages without any hanging with the same script .
Could you upload your test/ hanging script ?

cordovapolymer:
@GavinDmello the archive.zip in https://github.com/mcollina/aedes/issues/88 .
It doesn't make sense since it's 10 by default https://nodejs.org/api/events.html#events_eventemitter_defaultmaxlisteners . And you set it back to the default value.

PR #91, opened 2017-03-03
added a test with overlapping subscription

mcollina:
cc @gavindmello @yutter @wuhkuh 

This is a failing test.


mcollina:
The latest commit should fix it, thanks to @GavinDmello ^^.

PR #96, opened 2017-03-29
allow set server id for keep server id when restart

nguyenthenguyen:
When using `aedes-persistence-redis` and server restart, new server id generated, will topic has retain with old server id, server start `heartbeat` will publish will topic to client by https://github.com/mcollina/aedes/blob/master/aedes.js#L94


mcollina:
I'm a bit lost on why this will fix a will topic problem. Maybe we need to fix the will topic problem differently.

Anyhow, would you mind to add a unit test?

GavinDmello:
Maybe while streaming the wills we can replace the brokerId with the current broker Id ?

mcollina:
@GavinDmello yes! that's better, and it fixes the problem for every persistence.

GavinDmello:
@nguyenthenguyen You can include that in your PR then

nguyenthenguyen:
I'm just try aedes, I don't know how to fix this better. @GavinDmello Could you?

PR #105, opened 2017-04-13
Passing brokerId while deleting will

GavinDmello:


nguyenthenguyen:
Should we publish old will message ?
https://github.com/mcollina/aedes/blob/master/aedes.js#L98
In case, will topic will publish on first time server `_clearWillInterval` before old will deleted ?
https://github.com/mcollina/aedes/pull/98#issuecomment-291090278

GavinDmello:
@nguyenthenguyen Nice question. I feel it should be published because the clients did disconnect ungracefully and then reconnect when the broker restarted . @mcollina  What do you think ?

mcollina:
@nguyenthenguyen no, the replacement is just internal, not logical.

PR #115, opened 2017-06-07
add client#unsubscribe(subscriptions, [callback])

arden:
add client#unsubscribe(subscriptions, [callback])

### DIFF
--- a/lib/handlers/unsubscribe.js
+++ b/lib/handlers/unsubscribe.js
@@ -11,18 +11,24 @@ function UnsubscribeState (client, packet, finish, granted) {
 
 function handleUnsubscribe (client, packet, done) {
   var broker = client.broker
+  broker.persistence.removeSubscriptions(client, packet.unsubscriptions, function (err) {
+    if (err) {
+      return done(err)
+    }
 
-  if (packet.messageId) {
-    broker.persistence.removeSubscriptions(client, packet.unsubscriptions, function (err) {
-      if (err) {
-        return done(err)
-      }
-
-      actualUnsubscribe(client, packet, done)
-    })
-  } else {
     actualUnsubscribe(client, packet, done)
-  }
+  })
+  // if (packet.messageId) {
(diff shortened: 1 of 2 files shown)


lib/handlers/unsubscribe.js
```diff
@@ -11,18 +11,24 @@ function UnsubscribeState (client, packet, finish, granted) {
...
+  //   broker.persistence.removeSubscriptions(client, packet.unsubscriptions, function (err) {
+  //     if (err) {
+  //       return done(err)
+  //     }
+  //
+  //     actualUnsubscribe(client, packet, done)
+  //   })
+  // } else {
+  //   actualUnsubscribe(client, packet, done)
+  // }
```
mcollina: What is this breaking? I fear this is because how we handle the close pipeline.

mcollina:
Can you add unit test? Tests are failing as well.

lib/handlers/unsubscribe.js
arden: i'm run it on my application is well. 
i will test again.

GavinDmello: Is the packet.messageId check important? It's not there in the changes

mcollina: I think the check is there to handle some weird situation. It was probably a hack I put in :/.

arden: why in subscribe.js

```
function handleSubscribe (client, packet, done) {
  var broker = client.broker
  var subs = packet.subscriptions
  var granted = []

  broker._series(
    new SubscribeState(client, packet, done, granted),
    doSubscribe,
    subs,
    completeSubscribe)
}
```
haven't check the pack.messageId?

GavinDmello:
```
https://github.com/mcollina/aedes/blob/master/lib/handlers/unsubscribe.js#L15
```
You're calling the persistence removeSubs every time now which wasn't happening previously

mcollina:
Closing for no activity. Feel free to reopen in case you would like to keep working on it.

PR #121, opened 2017-07-01
Fix invalid subscription handling

behrad:
1. Fix `addSubscriptions` being called per topic, it should be called once for the topics array as said in README

2. Fix `addSubscriptions` being called when restoring client subs, it this case aedes-cached-`persistence#addedSubscriptions` should only be called.

3. Don't emit subscribe on restoring client subs.

fix number 1 and 2 are fast & dirty, but that not bad.
A better fix would need 

- changing the whole handleSubscribe, not to call `storeSubscriptions` per topic
- introducing a `restoreSubscriptions` method
- moving up (and renaming for a better one) `addedSubscriptions` up into aedes subscribe handler, I don't think [this](https://github.com/mcollina/aedes-cached-persistence/blob/master/index.js#L79-L93) is a implementation that should be inside aedes-cached-persistence

however these require a total refactoring in all persistence implementations, Some should help for such a change.

### DIFF
--- a/lib/handlers/subscribe.js
+++ b/lib/handlers/subscribe.js
@@ -60,19 +60,20 @@ function storeSubscriptions (sub, done) {
   var broker = client.broker
   var perst = broker.persistence
 
-  // if (!packet.messageId) {
-  //   // restore subscriptions
-  //   perst._addedSubscriptions(client, [sub])
-  //   return done(null, sub)
-  // }
-
   if (packet.subscriptions[0].topic !== sub.topic) {
-    // don't call addSubscriptions on each topic
+    // don't call addSubscriptions per topic,
+    // TODO change aedes subscribe handle, but this is a fast & dirty fix for now
+    return done(null, sub)
+  }
+
+  if (!packet.messageId) {
+    // this is a restoreSubscription, don't addSubscriptions!
+    // TODO change aedes subscribe handle, but this is a fast & dirty fix for now
+    perst._addedSubscriptions(client, packet.subscriptions)
     return done(null, sub)
   }
 
   perst.addSubscriptions(client, packet.subscriptions, function (err) {
-    // perst._addedSubscriptions(client, packet.subscriptions)
     done(err, sub)
   })
 }
--- a/lib/handlers/subscribe.js
+++ b/lib/handlers/subscribe.js
@@ -134,19 +134,15 @@ function completeSubscribe (err) {
     done()
   }
 
-  sendRetained()
-
-  function sendRetained () {
-    var persistence = broker.persistence
-    var topics = packet.subscriptions.map(function (s) { return s.topic })
-    var stream = persistence.createRetainedStreamCombi(topics)
-    stream.pipe(through.obj(function sendRetained (packet, enc, cb) {
-      packet = new Packet(packet)
-      // this should not be deduped
-      packet.brokerId = null
-      client.deliver0(packet, cb)
-    }))
-  }
+  var persistence = broker.persistence
+  var topics = packet.subscriptions.map(function (s) { return s.topic })
+  var stream = persistence.createRetainedStreamCombi(topics)
+  stream.pipe(through.obj(function sendRetained (packet, enc, cb) {
+    packet = new Packet(packet)
+    // this should not be deduped
+    packet.brokerId = null
+    client.deliver0(packet, cb)
+  }))
 }
 
 function SubAck (packet, granted) {
--- a/lib/handlers/subscribe.js
+++ b/lib/handlers/subscribe.js
@@ -103,29 +103,13 @@ function subTopic (sub, done) {
 
   if (!client.subscriptions[sub.topic]) {
     client.subscriptions[sub.topic] = new Subscription(sub.qos, func)
-    broker.subscribe(sub.topic, func, sendRetained)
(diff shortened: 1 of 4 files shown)


lib/handlers/subscribe.js
```diff
@@ -60,6 +60,21 @@ function storeSubscriptions (sub, done) {
...
+    // TODO change aedes subscribe handle, but this is a fast & dirty fix for now
+    return done(null, sub)
+  }
+
+  if (!packet.messageId) {
+    // this is a restoreSubscription, don't addSubscriptions!
+    // TODO change aedes subscribe handle, but this is a fast & dirty fix for now
+    if (perst._addedSubscriptions) {
+      perst._addedSubscriptions(client, packet.subscriptions)
+    }
```
mcollina: I'm -1 in introducing knowledge about aedes-cached-persistence. Is this truly neede, even in the proper fix?

behrad: I was -100 :))
I should've introduced a `subscriptionsRestored` or `subscriptionsAdded` hook inside aedes' subscribe handler. Then, aedes-cached-persistence would implement it via `_addedSubscriptions`. is it OK?

mcollina: It's not 100% clear to me how the code with `subscriptionRestored` and `subscriptionAdded` would play, but yes, go ahead. Any reference here to `_addedSubscriptions`  should disappear.

behrad: Problem is `persistence#addSubscriptions` should be called only when new subscribes arrives, but it is currently always called (also on restore subscriptions)
Another issue is that, any logic related to subscription handling (like any children override logics as inside `_addedSubscriptions`) should not go inside `addSubscriptions`, since in both cases (new sub and restore sub) children may decide to run that logics.

So, I've also another proposal @mcollina 
aedes subscribe handler emits a `clientSubscription` event, so that persistence children can trigger and call their logic there. in this case `_addedSubscriptions` would be called on that listener inside aedes-cached-persistence.

mcollina: The event will not work, as https://github.com/mcollina/aedes-cached-persistence/blob/master/index.js#L79-L107 needs a callback.

BTW, are you sure you need to call `_addedSubscriptions` at all? https://github.com/mcollina/aedes-cached-persistence/blob/master/index.js#L79-L107 is used to update the cache in the other peers, but as it is a restore, I do not think we need it.

mcollina:
Tests are failing badly with this change.

lib/handlers/subscribe.js
behrad: > as it is a restore, I do not think we need it

OMG! Yes, we can remove it at all!

behrad:
> Tests are failing badly with this change.

Found out `packet.messageId` is not a good guard to separate new/restore sub.
Since `client.subscribe` is not issued by a SUBSCRIBE but it must work.

mcollina:
Can you add a test for this, or it is just too complex? does it lead to improve perf?

behrad:
fixed @mcollina

behrad:
> Can you add a test for this, or it is just too complex? does it lead to improve perf?

1. a potential test can be addSubscriptions should be called once in a client.subscribe
2. another is `subscribe` event will only be called on client subscribes (not restores)
3. slight performance improvements

the notable one is the first, but I think it will be changed in future, so I didn't bother to write a test mean time.

mcollina:
I would like to have some checks that makes this hard to regress, as memory fades away.

Why do you think 1. will change soon? If we change that thing, we'll get that test updated.

behrad:
this is also using aedes-persistence 4.0.0, Is this now mergeable if you are pleased with `createRetainedStreamCombi` default implementation?

lib/handlers/subscribe.js
```diff
@@ -130,13 +124,29 @@ function completeSubscribe (err) {
...
+    var persistence = broker.persistence
+    var topics = packet.subscriptions.map(function (s) { return s.topic })
+    var stream = persistence.createRetainedStreamCombi(topics)
+    stream.pipe(through.obj(function sendRetained (packet, enc, cb) {
+      packet = new Packet(packet)
+      // this should not be deduped
+      packet.brokerId = null
+      client.deliver0(packet, cb)
+    }))
+  }
```
mcollina: this does not need to be a function anymore

behrad: removed sendRetained

behrad:
@mcollina ping

behrad:
@GavinDmello ping

GavinDmello:
@mcollina @behrad Sorry about the delay, I'm a bit held up.
These chanages look okay to me 👍

lib/handlers/subscribe.js
```diff
@@ -130,13 +124,25 @@ function completeSubscribe (err) {
...
 
   if (packet.messageId) {
     write(client, new SubAck(packet, granted), done)
   } else {
     done()
   }
+
+  var persistence = broker.persistence
+  var topics = packet.subscriptions.map(function (s) { return s.topic })
```
GavinDmello: One small change, This could be straight for loop

behrad: Done @GavinDmello 
Can you also check https://github.com/mcollina/aedes-persistence-redis/pull/31 ?

PR #130, opened 2017-07-14
Retain false when subscriber is active

GavinDmello:
Fix for <a> https://github.com/mcollina/aedes/issues/110 </a>

### DIFF
--- a/aedes.js
+++ b/aedes.js
@@ -146,9 +146,8 @@ function storeRetained (_, done) {
 }
 
 function emitPacket (_, done) {
-  var packet = Object.assign({}, this.packet)
-  packet.retain = false
-  this.broker.mq.emit(packet, done)
+  this.packet.retain = false
+  this.broker.mq.emit(this.packet, done)
 }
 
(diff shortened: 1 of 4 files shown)


mcollina:
Can you move the test to https://github.com/mcollina/aedes/blob/master/test/basic.js?

There is no need to use `Object.assign`, I think you can alter it in place.

GavinDmello:
@mcollina  There object is referenced from the _retained list in Aedes persistence.

mcollina:
I think we can create a new object in `Object.assign`  in https://github.com/mcollina/aedes-persistence/blob/master/persistence.js#L34 rather than creating a new one each time.

mcollina:
@GavinDmello the tests are failing on CI: https://travis-ci.org/mcollina/aedes/jobs/265538833#L1031-L1050

GavinDmello:
@mcollina  Yea. I'm checking. They seem to run locally though

mcollina:
I will land this with #131.

PR #136, opened 2017-07-21
Do not forward $SYS topics for +/# subscriptions.

mcollina:
Fixes: https://github.com/mcollina/aedes/issues/135

@ckrey

### DIFF
--- a/aedes.js
+++ b/aedes.js
@@ -200,8 +200,11 @@ function DoEnqueues () {
   }
 }
 
+// + is 43
+// # is 35
 function removeSharp (sub) {
-  return sub.topic !== '#'
+  var code = sub.topic.charCodeAt(0)
+  return code !== 43 && code !== 35
 }
 
 function doEnqueue (sub, done) {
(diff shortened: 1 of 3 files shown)


aedes.js
```diff
@@ -200,8 +200,12 @@ function DoEnqueues () {
   }
 }
 
+// + is 43
+// # is 35
 function removeSharp (sub) {
-  return sub.topic !== '#'
+  var code = sub.topic.charCodeAt(0)
+  console.log(sub.topic, code)
```
GavinDmello: remove console.log

GavinDmello:
This may affect the <a>https://github.com/mcollina/aedes/issues/123</a> change we were planning on implementing

mcollina:
@GavinDmello yes. Better to have the tests in.

aedes.js
mcollina: good catch!

mcollina: updated

ckrey:
✅

PR #140, opened 2017-07-21
Guarantee granted always come in the right order

mcollina:
Fixes https://github.com/mcollina/aedes/issues/133


mcollina:
@behrad I have reverted 607d99d6861d3d70beb88ef4dc7b5516e6f8d430.
So the tests here are passing. I'm releasing this as soon Travis is green. Feel free to send another PR that switches them back to parallel, but it does not break this additional test.

PR #158, opened 2017-10-05
clean clients should not have persistent subscription records.

sundryp:
If the current client is a clean session, a subscription should not be stored in persistence as it is not a persistent session.

### DIFF
--- a/lib/handlers/subscribe.js
+++ b/lib/handlers/subscribe.js
@@ -75,6 +75,10 @@ function storeSubscriptions (sub, done) {
     return done(null, sub)
   }
 
+  if (client.clean) {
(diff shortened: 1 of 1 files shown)


mcollina:
Can you add a unit test?

cordovapolymer:
@sundryp thanks for a patch, that's why redis and other persistences are [growing so big ](https://github.com/mcollina/aedes-persistence-redis/issues/45)

mcollina:
This would need a unit test.

cordovapolymer:
@mcollina, now I'm trying to test this, but I cannot find the issue, subscriptions for clean clients must be cleared somewhere else.

PR #160, opened 2017-10-08
authorizePublish gets called before authenticate finishes

jdiamond:
Hi @mcollina: what do you think of this test? I can try to fix it if it's legit and you have any hints on where I should look. Thanks.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -48,6 +48,10 @@ function Client (broker, conn) {
     var buf = empty
     var client = that
 
+    if (client._paused) {
+      return
+    }
+
     that._parsingBatch--
     if (that._parsingBatch <= 0) {
       that._parsingBatch = 0
@@ -61,6 +65,17 @@ function Client (broker, conn) {
   this._nextBatch = nextBatch
 
   nextBatch()
+
+  function pause () {
+    that._paused = true
+  }
+  this._pause = pause
+
+  function unpause () {
+    that._paused = false
+    nextBatch()
+  }
+  this._unpause = unpause
 
   conn.on('readable', nextBatch)
   conn.on('error', this.emit.bind(this, 'error'))
--- a/lib/client.js
+++ b/lib/client.js
@@ -66,17 +66,6 @@ function Client (broker, conn) {
 
   nextBatch()
 
-  function pause () {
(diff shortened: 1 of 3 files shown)


jdiamond:
My commit was a failing test to demonstrate the bug. Guess I should have expected Travis would fail when I had to bypass the pre-commit hooks...

mcollina:
@jdiamond I think we should somehow stop parsing of new packets while processing a CONNECT (see https://github.com/mcollina/aedes/blob/master/lib/client.js).

jdiamond:
Don't know if my commit is a good way to do the pausing, but it does make my test pass. I also pushed my branch to our QA environment where I first saw this bug and I'm not seeing it anymore.

Travis says there is a failing test in node 7. Not sure why, though. I just ran all tests 5 times in a row using node 7 and never saw any failures.

lib/client.js
```diff
@@ -62,6 +66,17 @@ function Client (broker, conn) {
 
   nextBatch()
 
+  function pause () {
+    that._paused = true
+  }
+  this._pause = pause
```
mcollina: can you define this a prototype method?

lib/client.js
```diff
@@ -62,6 +66,17 @@ function Client (broker, conn) {
...
+
+  function unpause () {
+    that._paused = false
+    nextBatch()
+  }
```
mcollina: this as well.

lib/handlers/connect.js
```diff
@@ -14,14 +14,26 @@ function ClientPacketStatus (client, packet) {
...
 
+function pauseParsing (arg, done) {
+  this.client._pause()
+  done()
+}
+
+function resumeParsing (arg, done) {
+  this.client._unpause()
+  done()
+}
```
mcollina: These are sync steps, I will try to embed them in others, i.e. authenticate and emptyQueue.

jdiamond:
Updated per your comments.

Same test is failing, but in node 4 this time. I still can't see how it's related to my change, though.

mcollina:
Thanks!!

jdiamond:
Thanks for the quick publish. Behaving great now. Next stop, production. 😄

PR #169, opened 2017-11-09
Just for review -)

syberkitten:
@mcollina 
This is pull request for review and related to our latest discussion.

Changed the tests to allow emitting('publish') with original retained flag (set by user when publishing, or broker when sending retained messages)

I've also added what is called an IPCClient - which is for our internal usages
and perhaps does not have place inside Aedes, probably should be thought of a bit more to conform into the Aedes eco-ssystem.

The IPCClient uses local socket to connect to an IPC process which is keeping
track of all connected / disconnected client from each process.

so in the case of an architecture having 10 nodejs processess each running an Aedes server,
the IPC will know which client is online and which is not.

the other parts of the IPC are the Near gateway and far gateway, both are part of a module called ClusterLink.

Again, this is all part of our development, and would love perhaps to see Aedes
grow into a Clustered solution for MQTT.

### DIFF
--- a/test/basic.js
+++ b/test/basic.js
@@ -163,7 +163,7 @@ test('live retain packets', function (t) {
     cmd: 'publish',
     topic: 'hello',
     payload: Buffer.from('world'),
-    retain: false,
+    retain: true,
     dup: false,
     length: 12,
     qos: 0
--- a/test/qos1.js
+++ b/test/qos1.js
@@ -427,7 +427,7 @@ test('deliver QoS 1 retained messages', function (t) {
     qos: 1,
     dup: false,
     length: 14,
-    retain: false
+    retain: true
   }
 
   subscribe(t, subscriber, 'hello', 1, function () {
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,7 @@ flamegraph.html
 .__*
 profile*
 
+.idea
+.DS_Store
+dump.rdb
+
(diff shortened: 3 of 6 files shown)


mcollina:
Thanks for contributing!

Can you remove IPCClient? I would like to keep this module as lean as possible. Feel free to add more events if you need them to support IPCclient as an external dependency!

test/qos1.js
```diff
@@ -427,7 +427,7 @@ test('deliver QoS 1 retained messages', function (t) {
     qos: 1,
     dup: false,
     length: 14,
-    retain: false
+    retain: true
```
mcollina: this should stay as it was. retain must be false here.

test/basic.js
```diff
@@ -163,7 +163,7 @@ test('live retain packets', function (t) {
     cmd: 'publish',
     topic: 'hello',
     payload: Buffer.from('world'),
-    retain: false,
+    retain: true,
```
mcollina: this must stay false

cordovapolymer:
@syberkitten , it would be great having IPCclient as a module for Aedes.

syberkitten:
@mcollina Thanks for the feedback..
since the client connect / disconnect events are dispatched rom the broker,
I've moved the IPC client out of Aedes, thus this pull request is closed.

more urgent issues right now are creating a Redis cluster and having
it work with Aedes.

PR #170, opened 2017-11-13
test for #158

cordovapolymer:
@mcollina 
Please see my test for #158 
It doesn't fail without the patch from #158 ..

Am I doing it right?


mcollina:
Subscriptions are cleared on connect https://github.com/cordovapolymer/aedes/blob/5a4210ad477bd2477be2748a19485a4e5f7a173b/lib/handlers/connect.js#L96-L99. You should put some assertions on the persistence itself if you want to check that they should not be stored.

cordovapolymer:
@mcollina 
I've added the assertion on persistence subscriptions, can you review?

test/basic.js
```diff
@@ -414,6 +414,38 @@ test('restore QoS 0 subscriptions not clean', function (t) {
   })
 })
 
+test('do not restore QoS 0 subscriptions when clean', function (t) {
+  var broker = aedes()
+  var publisher
+  var subscriber = connect(setup(broker), { clean: true, clientId: 'abcde' })
+
+  subscribe(t, subscriber, 'hello', 0, function () {
+    subscriber.inStream.end()
```
mcollina: you should check after this, not after reconnection.

mcollina:
Can you remove the yarn-lock file?

test/basic.js
cordovapolymer: thanks! corrected that, now it fails on a master without a patch from @sundryp

Issue #145, opened 2017-08-02
Emit 'unsubsribe' event when client has clean=true on disconnect


mcollina:
we should fire 'unsubscribe' for clean = true clients, but not for clean = false clients.

PR #171, opened 2017-11-15
Emit 'unsubsribe' event when client has clean=true on disconnect #145

cordovapolymer:
@mcollina , i've implemented #145, please review if it's ok to merge.

### DIFF
--- a/lib/handlers/unsubscribe.js
+++ b/lib/handlers/unsubscribe.js
@@ -68,7 +68,7 @@ function completeUnsubscribe (err) {
     return client.emit('error', err)
   }
 
-  if (!packet.close && packet.unsubscriptions.length > 0) {
+  if (client.clean === true && packet.unsubscriptions.length > 0) {
     client.broker.emit('unsubscribe', packet.unsubscriptions, client)
   }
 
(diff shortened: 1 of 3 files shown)


lib/handlers/unsubscribe.js
```diff
@@ -68,7 +68,7 @@ function completeUnsubscribe (err) {
     return client.emit('error', err)
   }
 
-  if (!packet.close && packet.unsubscriptions.length > 0) {
+  if (client.clean === true && packet.unsubscriptions.length > 0) {
```
mcollina: the check for `packet.close`  must stay. See https://github.com/mcollina/aedes/blob/275bda32461ea8053c8610b8a3e5b6f0f838f1b6/lib/client.js#L214.
Basically if `packet.close === true && client.clean === false` we should __not__ emit.

cordovapolymer: @mcollina , thank you for a clarification, please see the new commit

test/meta.js
```diff
@@ -123,7 +123,7 @@ test('emit unsubscribe event', function (t) {
   t.plan(6)
 
   var broker = aedes()
-  var s = connect(setup(broker), { clientId: 'abcde' })
+  var s = connect(setup(broker), { clean: true, clientId: 'abcde' })
```
mcollina: can you add another test that fails if the event is emitted when clean is false and the connection ends abruptly?

mcollina: foget about it, it's already added.

mcollina:
LGTM

PR #172, opened 2017-11-17
 modified tests and and benchmarks to test aedes with aedes-persistance-redis and mqemitter-redis

cordovapolymer:
@mcollina 

I have modified tests and and added a benchmark to test aedes with aedes-persistance-redis and mqemitter-redis.
The tests currently pass, and benchmarking server-redis_mqemitter-redis as I have [described](https://github.com/mcollina/aedes/issues/162#issuecomment-344674189) fails.


package.json.orig
```diff
@@ -0,0 +1,76 @@
...
+    "faucet": "0.0.1",
+    "istanbul": "^0.4.1",
+<<<<<<< HEAD
+    "mqemitter-redis": "^2.2.0",
+    "mqtt": "^2.13.0",
+    "mqtt-connection": "^3.0.0",
+=======
+    "mqtt": "^2.14.0",
+    "mqtt-connection": "^3.2.0",
+>>>>>>> 9a4eb0449d75a76db91099b32820f18f531a431b
```
nguyenthenguyen: Should we need this file and conflict code ?

cordovapolymer: no, thanks for noticing

mcollina:
I'm kind of 👎  on the test changes.

We need some integration tests between the various persistences, but these tests are designed to be _unit_, and I truly do not want any I/O to happen in there.

Not having an in-isolation test suite is one of the biggest issue in Mosca design, and I am not going through that route again.

Where you able to identify the bug?

cordovapolymer:
@mcollina , changed unit test didn't show any problems, you can reproduce the bug with benchmark as I have [described](https://github.com/mcollina/aedes/issues/162#issuecomment-344674189).

How do you see testing aedes with different persistences then? The only change I have made to the test system is that I've made the aedesConfig variable declarable to support other than the default persistence.

mcollina:
Running the multiple persistence tests in the Mosca repo is the source of flakyness in tests. Because after we have added redis, we have to add MongoDB, and then whatever other combo. This is not the place for the integration tests.

cordovapolymer:
@mcollina , so what would be your suggestion? Creating a separate repo with integration tests?

mcollina:
Yes! Really, spawning up a server (in a different process, ideally) and then calling it via MQTT.js, but test only basic cases.

PR #181, opened 2017-12-28
fixed Will message not removed from persistence after publish

emilianobonassi:
Fixes https://github.com/mcollina/aedes/issues/180


mcollina:
Thanks!!!

PR #183, opened 2018-01-22
Generic error event for missing client Ids

GavinDmello:


lib/client.js
```diff
@@ -151,7 +151,11 @@ function onError (err) {
   this.errored = true
   this.conn.removeAllListeners('error')
   this.conn.on('error', nop)
-  this.broker.emit('clientError', this, err)
+  if (this.id) {
+    this.broker.emit('clientError', this, err)
+  } else {
+    this.broker.emit('genericError', this, err)
```
mcollina: given that this can happen only during the initial handshake (CONNECT packet), can we name it `'connectError'`, `'connectionError'` or `'handshakeError'`? On second thoughts, this is not generic at all.

GavinDmello: Alright. I'm not sure why the tests are flaky

GavinDmello: I'm naming it `connectionError`

PR #196, opened 2018-05-11
Adding Typescript typings.

saboya:
Finishing the Typescript typings spree, following https://github.com/mqttjs/mqtt-packet/pull/39.

Unfortunately `aedes` needs export assignment, which limits the usage of the typinngs somewhat, but it's still very helpful.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,15 +3,21 @@
   "version": "0.33.0",
   "description": "Stream-based MQTT broker",
   "main": "aedes.js",
+  "types": "types/index.d.ts",
   "scripts": {
     "lint": "standard",
+    "tslint": "tslint types/**/*.d.ts",
+    "typescript-compile-test": "tsc -p test/typescript/tsconfig.json",
+    "typescript-compile-execute": "node test/typescript/*.js",
+    "typescript-test": "npm run typescript-compile-test && npm run typescript-compile-execute",
     "test": "tape test/*.js test/*/*.js | faucet",
-    "ci": "npm run lint; npm run coverage",
+    "ci": "npm run lint &&  npm run typescript-test && npm run coverage",
     "coverage": "istanbul cover tape test/*.js test/*/*.js",
     "coveralls": "cat coverage/lcov.info | coveralls"
   },
   "pre-commit": [
     "lint",
+    "tslint",
     "test"
   ],
   "repository": {
@@ -32,6 +38,7 @@
   "author": "Matteo Collina <hello@matteocollina.com>",
   "license": "MIT",
   "devDependencies": {
+    "@types/node": "^8.10.0",
     "compute-mode": "^1.0.0",
     "concat-stream": "^1.4.7",
     "convert-hrtime": "^2.0.0",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -47,6 +48,7 @@
     "websocket-stream": "^5.1.1"
   },
   "dependencies": {
+    "@types/node": "^8.10.0",
```
mcollina: I prefer `@types/node` to be a devDependency.

saboya: You're right, that's my bad.

saboya:
I adjusted the `@types/node` dependency and adjusted / added more typings. There are some issues with the typings though:

Like I said, `aedes` requires export asiggnment because it exports a function directly. This effectively prevents any other exports, which means that anything defined inside those typings cannot be referenced. This is not an issue for implicit typing, but prevents explicit typing.

Some cases where this might be an issue:

Return `Error` with `returnCode` set:
```typescript
import Aedes = require('aedes')

const authCallback: ,

const Aedes = Aedes({
  authCallback: (client, username: string, password: string, callback) => {
    if (username === 'test' && password === 'test') {
      callback(null, true)
    } else {
      // returnCode is not a property of Error, so this is the
      // only way to fulfill the desired interface from outside
      const error = new Error() as Error & { returnCode: number }
      error.returnCode = 1

      callback(error, false)
    }
  }
})
```

Indirect assignment of functions that use non-exported definitions:
```typescript
// Since we cannot reference Client our AuthCallback, we can't define them properly
const authCallback = (client: any, username: string, password: string, callback: any) => {
  if (username === 'test' && password === 'test') {
    callback(null, true)
  } else {
    const error = new Error() as Error & { returnCode: number }
    error.returnCode = 1

    callback(error, false)
  }
}

const Aedes = Aedes({
  authenticate: authCallback
})
```

Since we cannot re-export packet typings from `mqtt-packet` like it's done on `MQTT.js`, usage of those typings require explicit import by the user as well.

Those might or might not be deal-breakers.

types/index.d.ts
```diff
@@ -0,0 +1,82 @@
+/// <reference types="node" />
+
+import { IPublishPacket, ISubscribePacket, ISubscription, IUnsubscribePacket } from 'mqtt-packet'
+import { Socket } from 'net'
```
mcollina: This can be any Duplex.

saboya: Fixed.

package.json
```diff
@@ -3,15 +3,21 @@
...
+  "types": "types/index.d.ts",
   "scripts": {
     "lint": "standard",
+    "tslint": "tslint types/**/*.d.ts",
+    "typescript-compile-test": "tsc -p test/typescript/tsconfig.json",
+    "typescript-compile-execute": "node test/typescript/*.js",
+    "typescript-test": "npm run typescript-compile-test && npm run typescript-compile-execute",
     "test": "tape test/*.js test/*/*.js | faucet",
-    "ci": "npm run lint; npm run coverage",
+    "ci": "npm run lint;  npm run typescript-test; npm run coverage",
```
mcollina: Can you use `&&` instead of `;`?

saboya: Sure, I was just sticking to what was already in place :)

Issue #197, opened 2018-05-19
Argument of type '"publish"' is not assignable to parameter of type '"clientError" | "connectionError"'.


bsommardahl:
I'm using TypeScript and I'm getting the following errors when compiling:

```
Argument of type '"publish"' is not assignable to parameter of type '"clientError" | "connectionError"'.
```
and
```
Argument of type '"subscribe"' is not assignable to parameter of type '"clientError" | "connectionError"'.
```
I checked `index.d.ts` and sure enough, "publish" and "subscribe" are not defined as options. I'll try to fix this right now and submit a PR.

PR #198, opened 2018-05-19
Added missing events according to the docs. #197

bsommardahl:


mcollina:
Thanks! Can you add a unit test for this in https://github.com/mcollina/aedes/tree/master/test/typescript?

bsommardahl:
I added some console logs like I found in the file. Is there something in the build pipeline that actually converts those somehow to real unit tests?

mcollina:
no, we are just validating that typescript can compile this.

Issue #200, opened 2018-05-23
Negated subscription still delivered on retained message [question/discussion]


bjyurkovich:
This may be a misunderstanding on my part or just a stupid error, and would like some clarification before solidifying [this](https://github.com/bjyurkovich/aedes-authorization-plugin), so I wanted to bring this up as more of question/discussion than call out a bug from the start...

Given a server implementation that looks like this:

```
const AedesServer = require("aedes");
const config = require("./configuration");
const redis = require("mqemitter-redis");
const aedesPersistenceRedis = require("aedes-persistence-redis");
const mq = redis(config.persistence.redis);
const httpServer = require("http").createServer();
const ws = require("websocket-stream");

const aedes = AedesServer({
  mq,
  persistence: aedesPersistenceRedis({
    ...config.persistence.redis,
    db: config.persistence.redis.db + 1
  })
});

const mqttServer = require("net").createServer(aedes.handle);
ws.createServer({ server: httpServer }, aedes.handle);

aedes.authorizeSubscribe = (client, sub, callback) => {
  sub = null;
  callback(null, sub);
};

httpServer.listen(config.websocketPort, function() {
  console.log("WS broker listening on port", config.websocketPort);
});

mqttServer.listen(config.port, function() {
  console.log("MQTT broker listening on port", config.port);
});
```

I found found that if the subscription is negated as stated in the [docs](https://github.com/mcollina/aedes/#instanceauthorizesubscribeclient-pattern-doneerr-pattern) based on an authorizer pattern, then the client can still subscribe to the topic without an error (as expected), and does not receive an newly published messages.  This is good.

> For example based on the above server,  if `client1` first subscribes to topic `hello/1/world/2`, and `client2` publishes to `hello/1/world/2`, then `client1` does not receive the published message (as expected).

However, if a retained message is published to a topic before the subscription, then the (seemingly) negated subscription still delivers the retained message to the client on connect, but then the client does not receive any additional published messages while being connected to the topic (assuming that this is because the subscription was negated).  This is unexpected behavior.

> For example, if `client2` publishes a *retained* message to `hello/1/world/2`, and `client1` subscribes to `hello/1/world/2` (after the publication of the retained `client2` publishing), then `client1` still receives the retained message, but no more additionally (retained or not) published messages to topic `hello/1/world/2` while being subscribed.

[...]

PR #201, opened 2018-05-23
Fix for Retained message still delivered on negated subscription 

bjyurkovich:
Attempts to address #200 

I did not attempt to address the TODO [here](https://github.com/mcollina/aedes/blob/master/lib/handlers/subscribe.js#L34), and to really do this right, it appears as though we should likely address that first (not sure if they are connected to my simple hack fix, but from a cursory look, it appears as though my fix would (should?) be upended with that TODO completion).  Time is not on my side, so I did not attempt to reorganize that TODO.

I half-assed the tests (meaning I did not add any to cover this case).  Technically, from coverage purity perspective, there is still coverage for this, but not sure how we would want to test for retained functionality given the separate abstraction of the persistence plugins.  Open to suggestions here.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,20 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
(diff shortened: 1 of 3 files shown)


mcollina:
why this was closed?

bjyurkovich:
Styling was a mess...fixing and will redo the request.

PR #202, opened 2018-05-23
Fix/negated subscription retained delivery

bjyurkovich:
Hopefully for real this time (sorry for cluttering the repo).

Attempts to address #200

I did not attempt to address the TODO here, and to really do this right, it appears as though we should likely address that first (not sure if they are connected to my simple hack fix, but from a cursory look, it appears as though my fix would (should?) be upended with that TODO completion). Time is not on my side, so I did not attempt to reorganize that TODO.

I half-assed the tests (meaning I did not add any to cover this case). Technically, from coverage purity perspective, there is still coverage for this, but not sure how we would want to test for retained functionality given the separate abstraction of the persistence plugins. Open to suggestions here.


lib/handlers/subscribe.js
```diff
@@ -65,6 +62,13 @@ function storeSubscriptions (sub, done) {
   var broker = client.broker
   var perst = broker.persistence
 
+  // total hack. Love it. Good enough for government work.  Probably should be redone with that TODO down there...
+  if (!sub) {
+    client.isAuthorizedForSub = false
+  } else {
+    client.isAuthorizedForSub = true
+  }
```
mcollina: This should not work with multiple parallel subscribes, we probably need the proper fix for this.

bjyurkovich: I am seeing it now...yeah, this is going to require a lot more work :/

lib/handlers/subscribe.js
```diff
@@ -156,12 +160,16 @@ function completeSubscribe (err) {
     topics.push(packet.subscriptions[i].topic)
   }
   var stream = persistence.createRetainedStreamCombi(topics)
-  stream.pipe(through.obj(function sendRetained (packet, enc, cb) {
-    packet = new Packet(packet)
-    // this should not be deduped
-    packet.brokerId = null
-    client.deliver0(packet, cb)
-  }))
+  stream.pipe(
```
GavinDmello: We could probably not stream the message(s) by checking the granted QoS state, which would be 128 if negated.

bjyurkovich: I did not see the negated subscriptions getting a `[128]` granting...maybe I missed it?

bjyurkovich: It seems to me (now that I have more than 15 minutes of time with this codebase), that the best course of action would be what @GavinDmello mentioned:

Looking at:
```
  var persistence = broker.persistence
  var topics = []
  for (var i = 0; i < packet.subscriptions.length; i++) {
    topics.push(packet.subscriptions[i].topic) // <-- if we can push an object on here with granted status
  }
  var stream = persistence.createRetainedStreamCombi(topics) <-- this implements an interface
  stream.pipe(through.obj(function sendRetained (packet, enc, cb) {
    packet = new Packet(packet)
    // this should not be deduped
    packet.brokerId = null
    client.topics
    client.deliver0(packet, cb)
  }))
```
it seems like we can get the granted from the providing function.  The issue here is with scoping... by providing a streamed function `sendRetained` from the stream created by the `createRetainedStreamCombi` abstraction, we are somewhat limiting our ability to access the granting (without changing the persistance interface which is implemented by a number of other things like mongo and redis).  We possibly do some trickery, but it would be ugly...

bjyurkovich: Something like this?  Pretty rough hijacking `client` to do this...

```
var persistence = broker.persistence
  var topics = []
  client.granted = []
  for (var i = 0; i < packet.subscriptions.length; i++) {
    topics.push(packet.subscriptions[i].topic)
    client.topicAuthorization = {topic:packet.subscriptions[i].topic, granted:granted[i]}
  }
  var stream = persistence.createRetainedStreamCombi(topics)
  
  stream.pipe(through.obj(function sendRetained (packet, enc, cb) {
    packet = new Packet(packet)
    // this should not be deduped
    packet.brokerId = null
    for (var i = 0; i < client.topicAuthorization; i++) {
      if (client.topicAuthorization[i].topic === packet.topic 
          && client.topicAuthorization[i].granted 
          && client.topicAuthorization[i].granted !== 128){
          client.deliver0(packet, cb)
      }
    }
    
  }))
```

GavinDmello: @mcollina  @bjyurkovich Maybe we could and write the ack and terminate the flow before. I'm not sure if that's the best solution though.

bjyurkovich: @GavinDmello If you terminate the ack, doesn't that kick the subscription from the client?  Is that expected behavior as part of the 3.1.1 spec?

If that is a path forward, it looks like https://github.com/mcollina/aedes/blob/master/lib/handlers/subscribe.js#L89 could be a place to just remove the topic from `client` and simply don't register the granted status.  Is that what you mean?

Sorry, I may not be helpful...just peripherally looking when I get a few minutes here or there...

PR #204, opened 2018-06-06
emit client errors like unauthorized

### LINKED ISSUE #203: Authentication Client Errors
Aedes is not emitting mqtt connect errors like authentication errors: https://github.com/mcollina/aedes/blob/275bda32461ea8053c8610b8a3e5b6f0f838f1b6/lib/handlers/connect.js#L67-L83

How is this supposed to be done @mcollina ?
I'm suggesting emitting `clientError` with the specific error.

behrad:
this should fix #203

### DIFF
--- a/lib/handlers/connect.js
+++ b/lib/handlers/connect.js
@@ -60,28 +60,43 @@ function authenticate (arg, done) {
     negate)
 
   function negate (err, successful) {
+    var errorMessages = [
+      '',
+      'unacceptable protocol version',
+      'identifier rejected',
+      'Server unavailable',
+      'bad user name or password',
+      'not authorized'
+    ]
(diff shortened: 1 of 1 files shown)


lib/handlers/connect.js
```diff
@@ -60,28 +60,43 @@ function authenticate (arg, done) {
...
 
   function negate (err, successful) {
+    var errorMessages = [
+      '',
+      'unacceptable protocol version',
+      'identifier rejected',
+      'Server unavailable',
+      'bad user name or password',
+      'not authorized'
+    ]
```
mcollina: Can you move this to the top o the file?

behrad: done 👍

PR #206, opened 2018-06-22
Improve Typescript typings to allow explicit typing

rafsawicki:
With current type definitions it's not possible to use explicit typing. I changed them based on [this template](https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/declaration%20files/templates/module-function.d.ts.md) to allow to import all defined types.

Besides that, I added an `AuthenticateError` type and changed typescript tests to use explicit typing in a few places. I also changed some names in tests file to avoid naming collisions and to be more in line with other tests.

Not being able to use explicit typing is a problem in my project, so I hope you will find these changes useful.

### DIFF
--- a/test/typescript/typings.ts
+++ b/test/typescript/typings.ts
@@ -1,11 +1,10 @@
 // relative path uses package.json {"types":"types/index.d.ts", ...}
 
-import aedes = require ('../..')
-import { Client, AuthenticateError } from '../..'
+import { Server, Client, AuthenticateError } from '../..'
 import { IPublishPacket, ISubscribePacket, ISubscription, IUnsubscribePacket } from 'mqtt-packet'
 import { createServer } from 'net'
 
-const broker = aedes({
+const broker = Server({
   concurrency: 100,
   heartbeatInterval: 60000,
   connectTimeout: 30000,
--- a/test/typescript/typings.ts
+++ b/test/typescript/typings.ts
@@ -1,24 +1,25 @@
 // relative path uses package.json {"types":"types/index.d.ts", ...}
 
-import Aedes = require ('../..')
+import aedes = require ('../..')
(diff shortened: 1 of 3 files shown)


rafsawicki:
I don't think that my changes are cause of Travis build failing. It failed only on node 10 on some timeout related tests and I haven't touched the core logic at all.

test/typescript/typings.ts
```diff
@@ -1,24 +1,25 @@
 // relative path uses package.json {"types":"types/index.d.ts", ...}
 
-import Aedes = require ('../..')
+import aedes = require ('../..')
+import { Client, AuthenticateError } from '../..'
```
mcollina: I think we should really expose the server as `aedes.Server`.

Basically we should add to `aedes.js`:

``` js
aedes.Server = aedes
```

In this way you can do:

```js
import { Server, Client, AuthenticateError } from 'aedes'
```

This should not be a semver-major change, but just a facility for those using ESM and Typescript.

mcollina: I would prefer the way I suggested, as it’s an additive change and it does not require a change in ts configuration.

rafsawicki: I might have not explained it clearly - it wouldn't require ts configuration change, but change in the way it's currently imported. In any case, I added the `Server` property as you suggested.

sumitkm:
Is this PR waiting on something? Just readme update? I just switched from Mosca to Aedes and I find the TS definition currently released leaves a lot to be desired... this PR would do it a lot of good!

mcollina:
I'll do the quick readme update. Feel free to send more TS prs!

sumitkm:
@mcollina thanks muchly for merging the PR... I'll update my fork (which seems to be failing a test) and then have a quick run around the TS bindings again.

PR #212, opened 2018-08-06
Fix #211

### LINKED ISSUE #211: [SECURITY] Last will messages is not checked against authorization.
The LWT is not checked for authorization.

Simple example:
```js
const aedes = require('aedes')({ authorizePublish: (a, b, c) => c(new Error("Not allowed")) });
require('net').createServer(aedes.handle).listen(1883);
```

Now, from a client, connect and subscribe to "will".

From another client, connect with a last will message with topic "will". Then send a message to "test", breaking the connection (because of authorization failure).

Breaking the connection will trigger the LWT, and thus sends a message to "will", which can be seen in the other client.

mhverbakel:


lib/client.js
```diff
@@ -242,12 +242,20 @@ Client.prototype.close = function (done) {
...
+        function done (err) {
+          if (!err) {
+            that.broker.persistence.delWill({
+              id: that.id,
+              brokerId: that.broker.id
+            }, nop)
+          }
         }
       })
       that.will = null // this function might be called twice
```
nguyenthenguyen: Huhm! `authorizePublish` is async function. Server will crash when `that.broker.publish(that.will, that, done)` called because `that.will` is `null`

mhverbakel:
Correct, `that.will` should be replaced by a local copy.

Met vriendelijke groet,

Martijn Verbakel

nguyenthenguyen:
@mhverbakel @mcollina  Should we authorization will package when client connect?

mhverbakel:
I thought so too, but @mcollina thinks it should be on send (https://github.com/mcollina/aedes/issues/211#issuecomment-410800575)

PR #213, opened 2018-08-13
delivers a will waits for authorization

nguyenthenguyen:
Fixed for crash at [comment](https://github.com/mcollina/aedes/pull/212/files#diff-50cfa59973c04321b5da0c6da0fdf4feL253)

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -241,10 +241,11 @@ Client.prototype.close = function (done) {
   this._eos = nop
 
   function finish () {
-    if (!that.disconnected && that.will && !that._finished) {
-      that.broker.authorizePublish(that, that.will, function (err) {
+    if (!that.disconnected && that.will) {
+      var will = Object.assign({}, that.will)
+      that.broker.authorizePublish(that, will, function (err) {
         if (!err) {
-          that.broker.publish(that.will, that, done)
+          that.broker.publish(will, that, done)
         } else {
           done()
         }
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -251,14 +251,14 @@ Client.prototype.close = function (done) {
...
           if (!err) {
+            that.will = null // this function might be called twice
             that.broker.persistence.delWill({
               id: that.id,
               brokerId: that.broker.id
             }, nop)
           }
         }
       })
-      that.will = null // this function might be called twice
```
mhverbakel: Removing this line here allows the function to be called twice. You should leave this line, and make an in-closure variable for `that.will`.

lib/client.js
```diff
@@ -258,7 +258,7 @@ Client.prototype.close = function (done) {
...
+      that._finished = true // this function might be called twice
```
mcollina: a better solution is to extract `that.will` into a variable, and keep that into the context of the closure.

test/will.js
```diff
@@ -184,6 +184,28 @@ test('delivers a will with authorization', function (t) {
   s.conn.destroy()
 })
 
+test('delivers a will waits for authorization', function (t) {
+  let authorized = false
+  var opts = {}
+  // willConnect populates opts with a will
+  var s = willConnect(setup(aedes({ authorizePublish: (_1, _2, callback) => { authorized = true; setTimeout(() => { callback(null) }, 10) } })), opts)
```
mcollina: Can you shorten up the line and use a `setImmediate` instead?

lib/client.js
nguyenthenguyen: @mcollina Thanks you, sir!

PR #215, opened 2018-08-24
Removed a bad hack for older Node.js

mcollina:
This is not needed anymore.


GavinDmello:
I think we can merge this as we have already dropped support for everything under Node 4.

PR #217, opened 2018-10-11
allow broker id in options

jdiamond:
This allows `id` as an option when constructing a broker. When not defined, defaults to the old default of `shortid()`.


mcollina:
can you add a unit test?

Il giorno gio 11 ott 2018 alle 16:30 Jason Diamond <notifications@github.com>
ha scritto:

jdiamond:
Test added.

PR #220, opened 2018-10-16
remove xtend

### LINKED ISSUE #219: Use native Object.assign instead of xtend
[xtend](https://github.com/Raynos/xtend) can easily be replaced with [Object.assign](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign)

This reduce the dependencies and possible also avoid duplicated versions where some package don't use the same version range for the xtend package

```js
// immutable
Object.assign({}, a, b)

// mutable
Object.assign(a, b)
```

jimmywarting:
closes #219


mcollina:
I'm sorry but this now conflicts. Would you mind updating?

PR #227, opened 2019-01-20
test case for retained message on negated subsciption bug

GavinDmello:
Hey @mcollina 
This PR includes a failing test case for the bug mentioned in #200

### DIFF
--- a/test/client-pub-sub.js
+++ b/test/client-pub-sub.js
@@ -433,5 +433,5 @@ test('should not receive a message on negated subscription', function (t) {
     } else {
       t.pass('Message not received')
     }
-  }, 1000)
+  }, 100)
 })
--- a/test/client-pub-sub.js
+++ b/test/client-pub-sub.js
@@ -409,29 +409,29 @@ test('should not receive a message on negated subscription', function (t) {
       topic: 'hello',
       payload: Buffer.from('world'),
       qos: 0,
-      retain : true
+      retain: true
     }, function (err) {
+      t.error(err, 'no error')
+      client.subscribe({
+        topic: 'hello',
+        qos: 0
+      }, function (err) {
         t.error(err, 'no error')
-        client.subscribe({
-          topic: 'hello',
-          qos: 0
-        }, function (err) {
-          t.error(err, 'no error')
-        })
+      })
     })
   })
 
   var s = connect(setup(broker))
-  var receivedPacket = null;
+  var receivedPacket = null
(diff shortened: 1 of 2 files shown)


test/client-pub-sub.js
```diff
@@ -395,3 +395,43 @@ test('get message when client disconnects', function (t) {
...
+    receivedPacket = packet
+  })
+
+  setTimeout(function () {
+    if (receivedPacket != null) {
+      t.fail('Packet should not be received')
+    } else {
+      t.pass('Message not received')
+    }
+  }, 1000)
```
mcollina: Can you reduce this timer? Is there another test without a timer?

test/client-pub-sub.js
```diff
@@ -395,3 +395,43 @@ test('get message when client disconnects', function (t) {
...
+      }, function (err) {
+        t.error(err, 'no error')
+      })
+    })
+  })
+
+  var s = connect(setup(broker))
+  var receivedPacket = null
+  s.outStream.once('data', function (packet) {
+    receivedPacket = packet
```
mcollina: I think you can just call `t.fail()` here.

GavinDmello: The tests process would not exit using this approach as `t.fail()` will never be called and the plan takes 3 assertions.

GavinDmello:
@mcollina I've reduced the timer for now. Let me know if you have a better way.
I'll try to suggest a fix till then.

GavinDmello:
I noticed that we were calling addSubscription as well, which should not happen. Right now there are redundant checks in every series function which does not look good. Maybe we could break out of the series in this case ?

mcollina:
> I noticed that we were calling addSubscription as well, which should not happen. Right now the fix has redundant checks in every series function which does not look good. Maybe we could break out of the series in this case ?

I don't understand. However, +1 to the change.

GavinDmello:
There are redundant null checks in every series function with my approach. For eg https://github.com/mcollina/aedes/blob/0ea4a0c324f027165c2c2f834f943e08b018d89f/lib/handlers/subscribe.js#L68-L71  and https://github.com/mcollina/aedes/blob/0ea4a0c324f027165c2c2f834f943e08b018d89f/lib/handlers/subscribe.js#L97-L99. Ideally, I would have a check in the first series function and then complete the series somehow.

Anyway, we could merge this now and optimize later if required. 👍

Issue #208, opened 2018-07-17
Aedes publishes with QoS 2 when the client is subscribed with QoS 1


v-divya110:
Consider a client A which is subscribed to a topic T at QoS 1, and let client B publish a message to topic T at QoS 2, Aedes publishes the message to client A with QoS 2. 

Based on the MQTT spec: 
The subscription QoS gives the maximum QoS level at which the Server can send Application Messages to the Client."

3.8.4 provides non-normative examples here: http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718063

If a subscribing Client has been granted maximum QoS 1 for a particular Topic Filter, then a QoS 0 Application Message matching the filter is delivered to the Client at QoS 0. This means that at most one copy of the message is received by the Client. On the other hand a QoS 2 Message published to the same topic is downgraded by the Server to QoS 1 for delivery to the Client, so that Client might receive duplicate copies of the Message.

If the subscribing Client has been granted maximum QoS 0, then an Application Message originally published as QoS 2 might get lost on the hop to the Client, but the Server should never send a duplicate of that Message. A QoS 1 Message published to the same topic might either get lost or duplicated on its transmission to that Client.

**Non normative comment**

Subscribing to a Topic Filter at QoS 2 is equivalent to saying "I would like to receive Messages matching this filter at the QoS with which they were published". This means a publisher is responsible for determining the maximum QoS a Message can be delivered at, but a subscriber is able to require that the Server downgrades the QoS to one more suitable for its usage

mcollina:
This is a bug, as we are downgrading to QoS 0 effectively (see https://github.com/mcollina/aedes/blob/master/test/qos2.js#L180-L210). Would you like to send a PR?

Ostafen:
I have a similar problem. Messages delivered with QOS 2 are sent more than once, while QO1 behaves like QOS2

GavinDmello:
This bug should be solved with #229.

PR #229, opened 2019-02-02
Downgrade qos to client sub if needed

GavinDmello:
Solution for #208

### DIFF
--- a/test/qos2.js
+++ b/test/qos2.js
@@ -209,6 +209,39 @@ test('subscribe QoS 0, but publish QoS 2', function (t) {
   })
 })
 
+test('subscribe QoS 1, but publish QoS 2', function (t) {
+  var broker = aedes()
+  var publisher = connect(setup(broker))
+  var subscriber = connect(setup(broker))
+  var expected = {
+    cmd: 'publish',
+    topic: 'hello',
+    payload: Buffer.from('world'),
+    qos: 1,
+    dup: false,
+    length: 14,
+    retain: false
+  }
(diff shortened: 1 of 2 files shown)


test/qos2.js
```diff
@@ -209,6 +209,39 @@ test('subscribe QoS 0, but publish QoS 2', function (t) {
...
+    length: 14,
+    retain: false
+  }
+
+  subscribe(t, subscriber, 'hello', 1, function () {
+    subscriber.outStream.once('data', function (packet) {
+      delete packet.messageId
+      t.deepEqual(packet, expected, 'packet must match')
+      t.end()
+    })
```
mcollina: Can you add some checks on the follow up packets to verify that the full QoS flow is respected?

GavinDmello: Sure.

GavinDmello: @mcollina  I was just checking. You have helper functions in place to check the QoS flow.
 https://github.com/mcollina/aedes/blob/ca42eb0164e08538cabf2ddaa087ef1ff4b9dc8e/test/qos2.js#L11-L46 and https://github.com/mcollina/aedes/blob/ca42eb0164e08538cabf2ddaa087ef1ff4b9dc8e/test/helper.js#L77-L96

mcollina:
Yes we do! Can you use them in this test?

GavinDmello:
@mcollina I've already used the publish and subscribe helper functions in the test 👍.

PR #232, opened 2019-02-15
Updated deps

mcollina:
massive round of dependency updates


test/client-pub-sub.js
```diff
@@ -103,6 +103,7 @@ test('emit a `ack` event on PUBACK for QoS 1', function (t) {
   var s = connect(setup(broker))
 
   s.outStream.once('data', function (packet) {
+    console.log(packet)
```
GavinDmello: This log can be removed.

lib/client.js
```diff
@@ -305,7 +306,14 @@ Client.prototype.resume = function () {
 
 function enqueue (packet) {
   this.client._parsingBatch++
-  handle(this.client, packet, this.client._nextBatch)
+  // already connected or it's the first packet
+  if (this.client.connackSent || this.client._parsingBatch === 1) {
```
GavinDmello: Is it possible to use the `connackSent` event here instead of using a new event in `connected`?

mcollina: We can't. `'connackSent'`  is delivered too early, and due to the now-fixed asynchronicity of Readable, some packets might arrive early on. We need to wait for the full auth to complete, and that is async.

Note that this was *needed* also with previous versions, and it might be the cause of some obscure bugs/security vulnerabilities.

GavinDmello:
This looks good to go 👍

PR #233, opened 2019-02-16
callback instead of return in authorizeForward 

MrSpark2591:
Some async task can be performed now in authorizeForward before broker forwards this packet to consumer.
Let me know what you think and anything else needs to be changed. Thank you.
Note : This might be the breaking changes for the people who are already using this method by passing in options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,17 +286,16 @@ It will be called when a client is set to recieve a message. Override to supply 
 authorization logic.
 
 ```js
-instance.authorizeForward = function (client, packet) {
+instance.authorizeForward = function (client, packet, callback) {
   if (packet.topic === 'aaaa' && client.id === "I should not see this") {
-    return null
-    // also works with return undefined
+    callback()
   }
 
   if (packet.topic === 'bbb') {
     packet.payload = new Buffer('overwrite packet payload')
   }
 
-  return packet
+  callback(packet)
 }
 ```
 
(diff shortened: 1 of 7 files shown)


mcollina:
Thanks for the contribution, the code is great. However I cannot accept it as it is too much of a security risk.

`authorizeForward` was built as a synchronous function for a reason. A pub/sub broker is a fan-out component, a client publish one message and the broker delivers it to _N_ subscribers. If we enable `authorizeForward` to be asynchronous, we would be allowing a client to cause _N_ asynchronous operations. An asynchronous operation as memory requirements, and as such we can do only so many in parallel. Under high load, this feature could allow a client to cause an out of memory on the broker itself.

The correct approach to implement this feature is to fetch some rules for delivery on connect, and then use those rules to determine if we should forward a packet or not. This approach forces the use of an in-memory cache for the rules, so you'd have to keep them up-to-date somehow. A PR or a module with an example to implement all the above would be fantastic.

MrSpark2591:
Yes, i got your concern. Let me try out the approach you suggested.

MrSpark2591:
I have thought on your approach and as per your approach i don't think we need any further changes in the module.
Steps to achieve it:
1) We already have pre-connect method [`authenticate`](https://github.com/mcollina/aedes#instanceauthenticateclient-username-password-doneerr-successful) which can be used for getting stored data for the client from any persistence storage and put them into memory.
2) We can manipulate/use in memory client values in methods like `authorizePublish `, `authorizeSubscribe ` and  `authorizeForward ` as per our logic.
3) Once client gets disconnected we already have `clientDisconnect ` event if we want to update values for the client in persistence storage.

Let me know if i am missing something here. Thank you for the guidance.

PR #240, opened 2019-05-29
Fix bug while client subscribing to an Array of topics.

seeLuck:
1.fix bug while client(MQTT.js) subscribing to an Array of topics. I am curious the original line use only subscriptions[0] in if condition that would ignore the rest topics in an Array.

-2.fix bug while negating a subscription. The original logic is setting sub to null that will cause series problem. e.g: to negate multi topics and to persistent negated topics (which QoS is still 0 in storage).

-----------------

I retract #2 to pass test case with bugs I mentioned. Maybe you can change this part while you have time :)

--------------------------

I fixed negated topics persistence bug in the third commit.

Multi topics negating bug is still there.
```
instance.authorizeSubscribe = function (client, sub, callback) {
  if (sub.topic === 'aaaa') {
    sub = null
  }
  if (sub.topic === 'bbbb') { //bug here
    sub = null
  }
  callback(null, sub)
}
```
 I have tried to fix it but it seems to be related several parts and cannot pass test case.

I am not sure whether this is a graceful solution but it can fix bugs of my project for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -275,12 +275,12 @@ instance.authorizeSubscribe = function (client, sub, callback) {
 }
 ```
 
-To negate a subscription, set the subscription to `null`:
+To negate a subscription, set the subscription's QoS to `128`:
 
 ```js
 instance.authorizeSubscribe = function (client, sub, callback) {
   if (sub.topic === 'aaaa') {
-    sub = null
+    sub.qos = 128
   }
 
   callback(null, sub)
--- a/README.md
+++ b/README.md
@@ -275,12 +275,12 @@ instance.authorizeSubscribe = function (client, sub, callback) {
 }
 ```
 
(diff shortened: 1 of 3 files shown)


seeLuck:
> Would you mind adding a unit test?

Done.

I noticed an example of 'negate multiple subscriptions' in test case, code like
```
instance.authorizeSubscribe = function (client, sub, callback) {
  if (sub.topic === 'aaaa') {
       setTimeout(function () {
        cb(null, null)
      }, 100)
  }
  if (sub.topic === 'bbbb') { // is ok
       setTimeout(function () {
        cb(null, null)
      }, 50)
  }
  callback(null, sub)
}
```
to avoide the bug I mentioned before. However it looks a little bit strange to avoide them callback at the same time, I left this part in the original logic and only fixed bug about multi topics persistence.

PR #243, opened 2019-06-28
Gracefully Close mqemitter

gnought:


mcollina:
would you mind adding a unit test?

gnought:
@mcollina Added a unit test. My test goes fine but another "disconnect if a connect does not arrive in time" in test/keepalive.js goes fail. Please kinda pass this PR and see if you could fix the failed test.

PR #247, opened 2019-07-08
Emit PubAck, PubRec after publish in QoS 1 and 2

gnought:
PubAck and PubRec are responses based on MQTT 3.11 spec.


mcollina:
Can you please add a unit test?

gnought:
Tests fail in "subscribe and publish QoS 1 in parallel" https://github.com/mcollina/aedes/blob/master/test/qos1.js#L677, please see the comment in https://github.com/mcollina/aedes/issues/249

mcollina:
Can you fix the tests then?

gnought:
We need more code changes to fix the https://github.com/mcollina/aedes/issues/250 in QoS 2

gnought:
This PR not yet finish, we also need to fix QoS 2 packet order.

gnought:
This PR is not valid, something I have messed up, the received `publish` is from broker to sender, so the first *received* packet is `puback`

PR #251, opened 2019-07-11
Non-Zero MessageID when QoS > 0

gnought:
Aligned the MQTT 3.1.1 spec, http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718025


gnought:
The CI test will pass after the PR https://github.com/mcollina/aedes/pull/252 is applied

mcollina:
Would you mind to rebase?

PR #255, opened 2019-07-11
[MQTT-2.3.1-5] MUST not contain a packet identifier in QoS 0 PUBLISH packets

gnought:
This fix the issue https://github.com/mcollina/aedes/issues/254

### DIFF
--- a/aedes.js
+++ b/aedes.js
@@ -231,6 +231,8 @@ Aedes.prototype.publish = function (packet, client, done) {
   var publishFuncs = publishFuncsSimple
   if (p.qos > 0) {
     publishFuncs = publishFuncsQoS
+  } else {
+    delete p.messageId
   }
   this._series(new PublishState(this, client, p), publishFuncs, null, done)
 }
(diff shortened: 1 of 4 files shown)


mcollina:
Is a packet identifier being sent on the wire, or is this something that is just being added?

I think the mistake is in https://github.com/mcollina/aedes-packet/blob/6b3497ed99fca4ac6c7e308953516ee8d82f8b33/packet.js#L11. We should default that to `undefined`  or set it to `0`  depending on the QoS level.

gnought:
From the spec,

> Some types of MQTT Control Packets contain a variable header component. It resides between the fixed header and the payload. The content of the variable header varies depending on the Packet type. The Packet Identifier field of variable header is common in several packet types.

> The variable header component of many of the Control Packet types includes a 2 byte Packet Identifier field. These Control Packets are PUBLISH (where QoS > 0), PUBACK, PUBREC, PUBREL, PUBCOMP, SUBSCRIBE, SUBACK, UNSUBSCRIBE, UNSUBACK.

I think that the packet identifier is being added.
Sounds that it is good to make the `messageId` field be `undefined` as a default, but i don't know what the impact of packages depending on aedes-packet are.

gnought:
Maybe we could bump aedes-packet to 2.0.0 to minimise any impact.

gnought:
Submitted a PR https://github.com/mcollina/aedes-packet/pull/2, good enough?

PR #262, opened 2019-07-16
More reliable test in 'subscribe and publish QoS 1 in parallel'

gnought:
In a parallel testing, publish and suback may not arrive in order in different machines.

### DIFF
--- a/test/qos1.js
+++ b/test/qos1.js
@@ -686,6 +686,7 @@ test('subscribe and publish QoS 1 in parallel', function (t) {
     length: 14,
     retain: false
   }
+  var count = 0
 
   broker.on('clientError', function (client, err) {
     console.log(err.stack)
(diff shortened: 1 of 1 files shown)


test/qos1.js
```diff
@@ -695,19 +696,24 @@ test('subscribe and publish QoS 1 in parallel', function (t) {
...
+          cmd: 'puback',
+          messageId: packet.messageId
+        })
+        delete packet.messageId
+        t.deepEqual(packet, expected, 'packet must match')
+      }
+      if (count === 2) {
         t.end()
-      })
+      }
```
mcollina: Instead of using a count variable and t.end(), it’s better to use a t.plan() variable

gnought: thanks for your suggestion.

PR #282, opened 2019-07-23
Forcibly close a subscriber socket should not affect others

gnought:
When multiple subscribers listen a same topic, the other subscribers should still receive messages even one of them is destroyed. Refer to issue https://github.com/mcollina/aedes/issues/246


mcollina:
why you closed this?

gnought:
This PR is replaced by https://github.com/mcollina/aedes/pull/283

PR #283, opened 2019-07-25
Gracefully close clients & properly release connections

gnought:
This PR replaces PR https://github.com/mcollina/aedes/pull/282
and solves following issues
- https://github.com/mcollina/aedes/issues/237, Aedes not properly releasing connections
- https://github.com/mcollina/aedes/issues/246, Multiple clients subscribe same topic, no subscribers will receive message if one of them is closed by destroying socket

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -274,6 +274,9 @@ Client.prototype.close = function (done) {
     conn.removeAllListeners('error')
     conn.on('error', nop)
 
+    that.connected = false
+    that.disconnected = true
+
     if (that.broker.clients[that.id]) {
       that.broker.unregisterClient(that)
     }
@@ -289,9 +292,6 @@ Client.prototype.close = function (done) {
     } else {
       conn.end()
     }
-
-    that.connected = false
-    that.disconnected = true
 
     if (typeof done === 'function') {
       done()
--- a/lib/client.js
+++ b/lib/client.js
@@ -289,6 +289,9 @@ Client.prototype.close = function (done) {
       conn.end()
     }
 
+    that.connected = false
+    that.disconnected = true
+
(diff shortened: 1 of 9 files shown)


lib/client.js
```diff
@@ -273,6 +274,9 @@ Client.prototype.close = function (done) {
     conn.removeAllListeners('error')
     conn.on('error', nop)
 
+    that.connected = false
+    that.disconnected = true
```
mcollina: is disconnected a new property?

lib/write.js
```diff
@@ -3,12 +3,18 @@
...
-  var result = mqtt.writeToStream(packet, client.conn)
-
-  if (!result && !client.errored && done) {
-    client.conn.once('drain', done)
-  } else if (done) {
-    setImmediate(done)
+  if (client.conn.writable && client.connected) {
+    var result = mqtt.writeToStream(packet, client.conn)
+    if (!result && !client.errored && done) {
+      console.log('drain')
```
mcollina: I guess this is spurious

lib/client.js
gnought: `disconnected` is a existing property.
It is defined in https://github.com/mcollina/aedes/blob/master/lib/client.js#L46
and also be used in https://github.com/mcollina/aedes/blob/master/lib/handlers/index.js#L54

lib/write.js
gnought: you mean `if (client.conn.writable && client.connected) {`
or `if (!result && !client.errored && done) {` ?

mcollina: I mean the console.log.

gnought: ar. good catch. This is my debug purpose only. I will remove it. :)

PR #290, opened 2019-08-10
Enhance test & added test plan number

gnought:
For robustness, it is safer to state test plan number.


gnought:
@mcollina The upcoming PR mostly are enhancements/optimization, and some minor bug fixes except the attempt of fixing the maximum stack size exceeded.
We could do a release for let our users not to wait too long. :)

PR #294, opened 2019-08-11
Expose current Aedes version, used in aedes-stats

gnought:


mcollina:
Also, can you add a unit test?

gnought:
I think the unit test should be in aedes-stats project.
I will add the code
`
    doPub('version', aedesInstance.constructor.version)
` 
in aedes-stats `iterate()` function https://github.com/mcollina/aedes-stats/blob/master/stats.js#L52

mcollina:
This should just check that the version exposed is the same one of the
package.json.

Il giorno dom 11 ago 2019 alle 10:05 Gnought <notifications@github.com> ha
scritto:

gnought:
Add to prototype could be, but it will require an extra memory for storing version in each instance.
Version won’t change in each instance, except a user packs two aedes package. In this case he still can get different version using aedes.constructor.version, right?

gnought:
My idea is also be memory-wise

mcollina:
The following will not cause any addition memory allocation:

```js
Aedes.prototype.version =  require('./package').version
```

The reason for that is that it will stick the value on the prototype itself. Instances will look the value there, much like any additional function.

test/meta.js
```diff
@@ -229,6 +229,17 @@ test('emits client', function (t) {
   })
 })
 
+test('get aedes version', function (t) {
+  t.plan(0)
+
+  var broker = aedes()
+  if (broker.version === undefined) {
+    t.fail('version undefined')
```
mcollina: can you check if the content is the same of package.json?

PR #296, opened 2019-08-11
Drop safe-buffer

gnought:


gnought:
As we drop node 6 support in PR https://github.com/mcollina/aedes/pull/266

PR #297, opened 2019-08-11
Optimize handle function

gnought:


test/handlers/connect.js
```diff
@@ -13,12 +13,16 @@ test('reject clients with no clientId running on MQTT 3.1', function (t) {
   }
 
   client.broker = broker
+  client.conn = {
+    destroy: function () {}
```
mcollina: Why this change?

gnought: `handleConnect` is a sub handler in `handle`, and there's part of `connect` login in `handle` function. It is best to emulate much more real world case.

mcollina: If this is called, would you mind placing an assertion in there?

test/handlers/connect.js
```diff
@@ -13,12 +14,16 @@ test('reject clients with no clientId running on MQTT 3.1', function (t) {
   }
 
   client.broker = broker
+  client.conn = {
+    destroy: function () { assert.fail('should not destroy') }
```
mcollina: I think you can use t.fail() here instead

PR #301, opened 2019-08-13
Optimize Connect handler

gnought:
- For 06fc839, raise an "invalid protocol" callback error if there is

- aedes unsupported mqtt version: CONNACK return Code = 1
- client id length > 23 [MQTT 3.1.0 only]: CONNACK return Code = 2

- Added preConnect handler in handleConnect between earliest connect checks and normal checks. This is useful for users if they want to do some earilest DDoS check before server send any responses back, in this phrase connected=false

- Emit `connackSent` event and set connactSent=true if CONNACK is sent, not only after `client` event but also when normal connect checks phrase if necessary

- Added packet arguments in `connackSent` event

- not use uuid(), set clientID to 'aedes_' + shortid() if empty [MQTT 3.1.1], it is better to keep it within 23 chars for better compatibility

- Emit `clientReady` event after we send back all offline messages to client, for issue https://github.com/mcollina/aedes/issues/281

- Optimize negate function

- Optimize doConnack function and we could re-use it

- Set keepalive after authentication, save some resources if there are plenty of failed authentication

- Clear connect timer earliest in handleConnect before preConnect, performance-wise & memory-wise if there is DDoS


lib/handlers/connect.js
```diff
@@ -33,33 +40,44 @@ var errorMessages = [
...
   client._connectTimer = null
 
-  if (packet.keepalive > 0) {
-    client._keepaliveInterval = (packet.keepalive * 1500) + 1
-    client._keepaliveTimer = retimer(function keepaliveTimeout () {
-      client.broker.emit('keepaliveTimeout', client)
-      client.emit('error', new Error('keep alive timeout'))
-    }, client._keepaliveInterval)
+  if (client.broker.preConnect(client) === false) {
+    return client.conn.destroy()
```
mcollina: please do this in two statements, as having multiple returns types make the V8 optimizer sweat a bit more.

lib/handlers/connect.js
```diff
@@ -33,33 +40,44 @@ var errorMessages = [
...
+    returnCode = 2
+  }
+  if (returnCode > 0) {
+    var error = new Error(errorMessages[returnCode])
+    error.errorCode = returnCode
+    client.broker.emit('clientError', client, error)
+    doConnack(
+      { client: client, returnCode: returnCode, sessionPresent: false },
+      done.bind(this, error))
+    return client.conn.end()
```
mcollina: same here

README.md
```diff
@@ -213,6 +219,26 @@ Both `topic` and `payload` can be `Buffer` objects instead of strings.
...
 
+-------------------------------------------------------
+<a name="preConnect"></a>
+### instance.preConnect(client)
+
+It will be called when aedes instance receives a first valid CONNECT packet from client. client object state is in default and its connected state is false. Any values in CONNECT packet (like clientId, clean flag, keepalive) will pass to client object after this call. Override to supply custom preConnect logic.
+Some use cases:
+1. Rate Limit / Throttle by `client.conn.remoteAddress`
+2. Check `instance.connectedClient` to limit maximum connections
+3. IP blacklisting
```
mcollina: Given this list of use cases, I think preConnect should be asynchronous/have a callback. We might want to do rate limiting using a redis store for example.

Issue #281, opened 2019-07-23
Duplicate PUBLISH packets in directly push to a QoS 2 client.


gnought:
It seems that we get a serious bug.

When we direct push a PUBLISH to a QoS 2 client, we get duplicate publish packets. The following test is trying to demonstrate it. The `publishCount` is 2, and `nonPublishCount` is 1.
However the correct result should be `publishCount` is 1, and `nonPublishCount` is 1.

Test code:
```
test('publish direct to a single client QoS 2', function (t) {
  var broker = aedes()
  var messageId
  var clientId
  var publishCount = 0
  var nonPublishCount = 0

  broker.on('client', function (client) {
    clientId = client.id
    client.publish({
      topic: 'hello',
      payload: Buffer.from('world'),
      qos: 2
    }, function (err) {
      t.error(err, 'no error')
    })
  })

  var s = connect(setup(broker))

  s.outStream.on('data', function (packet) {
    if (packet.cmd === 'publish') {
      publishCount++
      console.log('s.outStream', packet)
      s.inStream.write({
        cmd: 'pubrec',
        messageId: packet.messageId
      })
    } else {
      nonPublishCount++
      messageId = packet.messageId
      s.inStream.write({
        cmd: 'pubcomp',
        messageId: packet.messageId
      })
    }
  })

  broker.on('closed', function () {
    console.log(publishCount)
    console.log(nonPublishCount)
  })
})
```

mcollina:
Is this against master or against the last published version? In other terms, is this the result of some of the changes to persistence? Would you mind sending a PR to fix?

gnought:
The test is based on the `master` branch with v6.0.0 persistence.

It seems that I find the root clause.
The publish packet is QoS 2 and will sit in offline queue until broker receives pubcomp.
The connect actions are 
```
var connectActions = [
  authenticate,
  fetchSubs,
  restoreSubs,
  storeWill,
  registerClient,
  doConnack,
  emptyQueue
]
```

1. `client` event is emitted when in `registerClient` action.
2. publish packet is then triggered to push it into the stream.
3. After some times, `emptyQueue` run and push the publish packet in offline queue again.

It makes sense that `client` event is raised when a new client connects, but it is too earlier for the test.
Similarly 2 tests also depends on `client` event.
https://github.com/mcollina/aedes/blob/master/test/client-pub-sub.js#L24
https://github.com/mcollina/aedes/blob/master/test/client-pub-sub.js#L55

[...]

Issue #257, opened 2019-07-12
Packet in Broker `ack` event


gnought:
In https://github.com/mcollina/aedes/blob/master/lib/handlers/puback.js#L6

Broker receives puback or pubcomp packet will go into puback.js

In non-clean QoS 1, the `origPacket` will be a PUBLISH packet
In non-clean QoS 2, the `origPacket` will be a PUBREL packet.

> ack: when a packet published to a client is delivered successfully with QoS 1 or QoS 2

A little bit confusion, should the `packet` in `ack` event always be the PUBLISH one?

In a clean QoS 1 and QoS 2, the `origPacket` will be `undefined`. That means we will not get any packet info in `ack` event, is it by design?

mcollina:
Can you add a code example to explain? I wrote this a long ago, and I do not remember / or I have time to figure this out.

gnought:
```
TAP version 13
# subscribe & publish QoS 1 with clean=true
ok 1 should be equal
ok 2 should be equivalent
ok 3 should be equal
QoS 1 [clean=true] ack undefined

# subscribe & publish QoS 2 with clean=true
ok 4 should be equal
ok 5 should be equivalent
ok 6 should be equal
QoS 2 [clean=true] ack undefined

# subscribe & publish QoS 1 with clean=false
ok 7 should be equal
ok 8 should be equivalent
ok 9 should be equal
QoS 1 [clean=false] ack PubRel { cmd: 'pubrel', messageId: 52676 }

# subscribe & publish QoS 2 with clean=false
ok 10 should be equal
ok 11 should be equivalent
ok 12 should be equal
QoS 2 [clean=false] ack PubRel { cmd: 'pubrel', messageId: 11823 }

1..12
# tests 12
# pass  12

# ok
```


The code
```
'use strict'

var Packet = require('aedes-packet')
var Buffer = require('safe-buffer').Buffer
var test = require('tape').test
var helper = require('./helper')
var aedes = require('../')
var setup = helper.setup
var connect = helper.connect
var subscribe = helper.subscribe

function publish (t, s, packet, done) {
  var msgId = packet.messageId

  s.inStream.write(packet)

  s.outStream.once('data', function (packet) {
    s.inStream.write({
      cmd: 'pubrel',
      messageId: msgId
    })
    s.outStream.once('data', function (packet) {
      if (done) {
        done()
      }
    })
  })
}

function receive (t, subscriber, expected, done) {
  subscriber.outStream.once('data', function (packet) {
[...]

PR #305, opened 2019-08-24
Enhance cilent pub sub tests

gnought:
Fixed tests that will produce duplicate PUBLISH packets in directly push to a QoS 2 client. #281
and state packet type in Broker `ack` event #257


gnought:
Replaced PR https://github.com/mcollina/aedes/pull/273

PR #308, opened 2019-08-24
Fixed PreConnect typescript definition

gnought:


mcollina:
Could you add a test for this?

PR #313, opened 2019-09-01
Dedupe subscriptions

gnought:
If there are duplicated topics in subscriptions, no matter what QoS is, we pick the last one.
say the subscriptions in SUBSCRIBE are
```
[{ topic: 'hello', qos: 1 }, { topic: 'hello', qos: 0 }]
```
the actual subscription is only `{ topic: 'hello', qos: 0 }`

Vernemq and mosquitto are in same behaviour.

### DIFF
--- a/lib/handlers/subscribe.js
+++ b/lib/handlers/subscribe.js
@@ -40,7 +40,7 @@ function handleSubscribe (client, packet, done) {
   broker._series(
     new SubscribeState(client, packet, done, granted),
     doSubscribe,
-    subs.length === 0 ? subs : _dedupe(subs),
+    subs.length === 1 ? subs : _dedupe(subs),
     completeSubscribe)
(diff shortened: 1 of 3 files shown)


lib/handlers/subscribe.js
```diff
@@ -19,6 +19,19 @@ function SubscribeState (client, packet, finish, granted) {
   this.subsIndex = 0
 }
 
+// if same subscribed topic in subs array, we pick up the last one
+function _dedupe (subs) {
+  var dedupeSubs = {}
+  subs.forEach(function (el) {
+    dedupeSubs[el.topic] = el
+  })
```
mcollina: Can we use  `for(;;)` loop instead?

gnought: sure

PR #315, opened 2019-09-01
Added empty subscriptions test

gnought:


test/basic.js
```diff
@@ -65,6 +65,24 @@ test('subscribe QoS 0', function (t) {
...
+    cmd: 'subscribe',
+    messageId: 24,
+    subscriptions: []
+  })
+
+  s.outStream.once('data', function (packet) {
+    t.fail('message received')
+  })
+  broker.on('closed', t.end.bind(t))
+})
```
mcollina: Are you sure this test is actually running? Who will emit ‘closed’?

gnought: The closed event is from https://github.com/mcollina/aedes/blob/master/test/helper.js#L22

gnought: Tried the following, we will receive `broker closed` message.
```
test('test empty subscriptions', function (t) {
  t.plan(0)

  var broker = aedes()
  var s = connect(setup(broker))

  s.inStream.write({
    cmd: 'subscribe',
    messageId: 24,
    subscriptions: []
  })

  s.outStream.once('data', function (packet) {
    t.fail('message received')
  })
  broker.on('closed', function () {
    console.log('broker closed')
    t.end()
  })
})
```

mcollina: I would prefer to not rely on a `setTimeout`. Would you mind restructuring the test in that way? A good way would be to send a DISCONNECT packet or similar, and check if the number of packet do not inclue a SUBACK.

BTW, it this behavior spec-compliant?

robertsLando:
@gnought @mcollina Could this be merged or closed?

gnought:
I have less time to check if it met the spec.
@robertsLando could you help to have a check? or test again mosquitto behaviour?

robertsLando:
@gnought Actually I don't understand where the `closed` event is emitted. Are you sure the test is still running?

robertsLando:
Based on [this](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718063):

> 3.8.3 Payload [...] The payload of a SUBSCRIBE packet MUST contain at least one Topic Filter / QoS pair. A SUBSCRIBE packet with no payload is a protocol violation [MQTT-3.8.3-3]. See section 4.8 for information about handling errors.

gnought:
@robertsLando the closed event is emitted at https://github.com/moscajs/aedes/blob/master/aedes.js#L300

Thanks for checking the spec. We need to rewrite the test to fit into @mcollina expected

robertsLando:
Yes but my question is who calls broker.close in your test? 

Daniel

gnought:
@robertsLando the setup() function will call broker.close in time-based. Check this out https://github.com/moscajs/aedes/blob/22748026f2c866a8b3dbb4a69b29dbe781017b1b/test/helper.js#L22 .

However it has already removed after when we switch from tape to tap, and we use teadown to properly close broker.

robertsLando:
Ok that is why I don’t see it on my master. 

Daniel

robertsLando:
@gnought I think we can close this as empty subscription is already tested here: https://github.com/moscajs/aedes/blob/master/test/topics.js#L96

gnought:
Yes, would you mind to add a comment in https://github.com/moscajs/aedes/blob/a61f8999c59ed78c5544a9d978a50e03c9177402/lib/handlers/validations.js#L7 ?
Like `// [MQTT-3.8.3-3]`

robertsLando:
@gnought done https://github.com/moscajs/aedes/pull/439

PR #318, opened 2019-09-02
Use uuid for broker id while shortid for client id

gnought:


aedes.js
```diff
@@ -36,12 +36,13 @@ function Aedes (opts) {
 
   opts = Object.assign({}, defaultOptions, opts)
 
-  this.id = opts.id || shortid()
+  this.id = opts.id || uuidv5('https://github.com/mcollina/aedes', uuidv5.URL)
```
mcollina: will this be different at every run? That's the reason why we use short-ids in the beginning.

gnought: Yes, the id will be different at every run. The parameter in uuidv5 is a namespace.
https://www.npmjs.com/package/uuid

gnought:
@mcollina sorry my bad, we need to change uuid/v5 to uuid/v4 for a different broker id at every run.

mcollina:
can you send another PR?

Il giorno lun 2 set 2019 alle ore 17:53 Gnought <notifications@github.com>
ha scritto:

PR #322, opened 2019-09-09
Add websocket/http request to client object during connection

abachman:
Servers based on websocket-stream require access to the original HTTP request headers in order to access the `X-Forwarded-For` or any other interesting HTTP header. 

I've done the minimal possible work to attach the original HTTP request to aedes' Client object so that it's available later on.

### DIFF
--- a/test/connect.js
+++ b/test/connect.js
@@ -5,6 +5,9 @@ var helper = require('./helper')
 var aedes = require('../')
 var setup = helper.setup
 var connect = helper.connect
+var http = require('http')
+var ws = require('websocket-stream')
+var mqtt = require('mqtt')
 
 ;[{ ver: 3, id: 'MQIsdp' }, { ver: 4, id: 'MQTT' }].forEach(function (ele) {
   test('connect and connack (minimal)', function (t) {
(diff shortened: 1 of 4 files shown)


mcollina:
This would need docs and a test, otherwise I’m good to add this.

abachman:
Thanks for the quick attention!

I added a websocket-stream based server + client test modeled after similar tests in `tests/not-blocking.js`. It creates the http server, connects a client with headers, and then checks to see that the headers are present. 

I also added brief notes to README.md on the `conn` and `req` properties of `Client` objects.

gnought:
LGTM

test/connect.js
```diff
@@ -361,3 +364,48 @@ test('reject clients with wrong protocol name', function (t) {
...
+
+  var client = mqtt.connect('ws://localhost:4883', {
+    wsOptions: {
+      headers: {
+        'X-Test-Protocol': 'sample'
+      }
+    }
+  })
+
+  var timer = setTimeout(finish, 1000)
```
gnought: Could we eliminate the `setTimeout` ?

abachman: If you feel it's necessary. It's copied from the same pattern in https://github.com/mcollina/aedes/blob/master/test/not-blocking.js#L18 as a clean up for failing tests.

gnought:
LGTM

CyanoFresh:
please make a release with this feature

PR #326, opened 2019-09-14
Fixed duplicated done call in ping handler

gnought:


gnought:
We already have ping tests in https://github.com/mcollina/aedes/blob/master/test/keep-alive.js, not sufficient?

mcollina:
If this fixes a bug, then the test was not covering that case. In order to avoid future regressions, we got to add tests.

robertsLando:
@mcollina @gnought Is there a way to create the test needed? I can't figure out a way to test the double call on handelPing. This may fix a severe bug related to clients status that I can't debug right now

mcollina:
Have you got a side effect that you have noticed?

robertsLando:
@mcollina #331 . In poor words there are problems with clients connection status. I get the client disconnected event even if the client is still connected (it maybe disconnect and reconnect too quickly that the connect event hasn't been sent?) also I received a will packet sent from broker even if the client was still connected. I'm sure about this bug, I'm not sure what may cause this but the ping could be a cause (?)

gnought:
I haven't much time to work on the test. Thanks all. :)

PR #334, opened 2019-11-26
Proxy and ip decoder

getlarge:
Hey,
I'm using this project very often and i noticed that having a nice ipAddress property might be useful for me and other users ( to use in preConnect hook for example ).
Since it's important to have a correct information, i thought it would be interesting to add a trustProxy property to the broker. Then if aedes is placed behind load balancer / proxy,  we could use http headers (x-real-ip / x-forwarded-for ) in case of websocket connection or proxy protocol in case of tcp socket.
So i implemented proxy-protocol-js to test incoming client's buffer and to try retrieving source ip address.

I've been adding an example and some tests.

Please let me know if this feature is useful and not out of the scope of aedes.

### DIFF
--- a/examples/proxy/index.js
+++ b/examples/proxy/index.js
@@ -7,6 +7,8 @@ var mqttPacket = require('mqtt-packet')
 var net = require('net')
 var proxyProtocol = require('proxy-protocol-js')
 
+var brokerPort = 4883
+
 function sendProxyPacket (version = 1) {
   var packet = {
     cmd: 'connect',
@@ -16,11 +18,11 @@ function sendProxyPacket (version = 1) {
     clientId: `my-client-${version}`,
     keepalive: 0
   }
-
+  var clientIp = '192.168.0.12'
   var protocol
   if (version === 1) {
-    var src = new proxyProtocol.Peer('127.0.0.1', 12345)
-    var dst = new proxyProtocol.Peer('127.0.0.1', 1883)
+    var src = new proxyProtocol.Peer(clientIp, 12345)
+    var dst = new proxyProtocol.Peer('127.0.0.1', brokerPort)
     protocol = new proxyProtocol.V1BinaryProxyProtocol(
       proxyProtocol.INETProtocol.TCP4,
       src,
@@ -32,10 +34,10 @@ function sendProxyPacket (version = 1) {
       proxyProtocol.Command.LOCAL,
       proxyProtocol.TransportProtocol.DGRAM,
       new proxyProtocol.IPv4ProxyAddress(
-        proxyProtocol.IPv4Address.createFrom([127, 0, 0, 1]),
+        proxyProtocol.IPv4Address.createFrom(clientIp.split('.')),
         12346,
         proxyProtocol.IPv4Address.createFrom([127, 0, 0, 1]),
-        1883
+        brokerPort
       ),
       mqttPacket.generate(packet)
     ).build()
@@ -72,15 +74,11 @@ function sendProxyPacket (version = 1) {
 
   mqttConn.on('timeout', function () {
     // console.log("protocol proxy buffer", data)
-    mqttConn.write(data, () => {
-      mqttConn.end()
-    })
+    mqttConn.end(data)
   })
 }
 
 function startAedes () {
-  var port = 1883
-
   var broker = aedes({
     mq: mqemitter({
       concurrency: 100
@@ -99,8 +97,8 @@ function startAedes () {
 
   var server = require('net').createServer(broker.handle)
 
-  server.listen(port, function () {
-    console.log('Aedes listening on port:', port)
+  server.listen(brokerPort, function () {
+    console.log('Aedes listening on port:', brokerPort)
     broker.publish({ topic: 'aedes/hello', payload: "I'm broker " + broker.id })
     setTimeout(() => sendProxyPacket(1), 250)
     setTimeout(() => sendProxyPacket(2), 500)
--- a/examples/proxy/index.js
+++ b/examples/proxy/index.js
@@ -1,102 +1,130 @@
+var aedes = require('../../aedes')
 var mqemitter = require('mqemitter')
 var persistence = require('aedes-persistence')
 var mqttPacket = require('mqtt-packet')
 var net = require('net')
 var proxyProtocol = require('proxy-protocol-js')
 
-function sendProxyPacket () {
+function sendProxyPacket (version = 1) {
   var packet = {
     cmd: 'connect',
     protocolId: 'MQTT',
     protocolVersion: 4,
     clean: true,
-    clientId: 'my-client',
+    clientId: `my-client-${version}`,
     keepalive: 0
   }
 
-  var data = mqttPacket.generate(packet)
-  console.log(data)
+  var protocol
+  if (version === 1) {
+    var src = new proxyProtocol.Peer('127.0.0.1', 12345)
+    var dst = new proxyProtocol.Peer('127.0.0.1', 1883)
+    protocol = new proxyProtocol.V1BinaryProxyProtocol(
+      proxyProtocol.INETProtocol.TCP4,
+      src,
+      dst,
+      mqttPacket.generate(packet)
+    ).build()
+  } else if (version === 2) {
+    protocol = new proxyProtocol.V2ProxyProtocol(
+      proxyProtocol.Command.LOCAL,
+      proxyProtocol.TransportProtocol.DGRAM,
+      new proxyProtocol.IPv4ProxyAddress(
+        proxyProtocol.IPv4Address.createFrom([127, 0, 0, 1]),
+        12346,
+        proxyProtocol.IPv4Address.createFrom([127, 0, 0, 1]),
+        1883
+      ),
+      mqttPacket.generate(packet)
+    ).build()
+  }
 
-  var src = new proxyProtocol.Peer('127.0.0.1', 12345)
-  var dst = new proxyProtocol.Peer('127.0.0.1', 1883)
-  var protocol = new proxyProtocol.V1BinaryProxyProtocol(
-    proxyProtocol.INETProtocol.TCP4,
-    src,
-    dst,
-    data
-  ).build()
+  var parsedProto = version === 1
+    ? proxyProtocol.V1BinaryProxyProtocol.parse(protocol)
+    : proxyProtocol.V2ProxyProtocol.parse(protocol)
+  // console.log(parsedProto)
 
-  console.log(protocol)
+  var dstPort = version === 1
+    ? parsedProto.destination.port
+    : parsedProto.proxyAddress.destinationPort
 
-  var parsedProto = proxyProtocol.V1BinaryProxyProtocol.parse(protocol)
-  console.log(parsedProto)
+  var dstHost = version === 1
+    ? parsedProto.destination.ipAddress
+    : parsedProto.proxyAddress.destinationAddress.address.join('.')
 
+  // console.log('Connection to :', dstHost, dstPort)
   var mqttConn = net.createConnection(
     {
-      port: parsedProto.destination.port,
-      host: parsedProto.destination.ipAddress,
-      // localAddress: proto.source.ipAddress,
-      // localPort: proto.source.port,
-      // family: parsedProto.ipFamily,
-      // lookup,
+      port: dstPort,
+      host: dstHost,
       timeout: 300
     }
   )
 
+  var data
+  if (version === 2) {
+    data = Buffer.from(protocol.buffer)
+  } else {
+    data = protocol
+  }
+
   mqttConn.on('timeout', function () {
-    mqttConn.write(protocol)
-    mqttConn.end()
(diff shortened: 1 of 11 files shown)


examples/proxy/index.js
```diff
@@ -0,0 +1,131 @@
+var aedes = require('../../aedes')
```
mcollina: Can you add "use strict" at the top?

lib/client.js
```diff
@@ -64,8 +146,12 @@ function Client (broker, conn, req) {
       that._parsingBatch = 0
       var buf = empty
       buf = client.conn.read(null)
-
-      if (buf) {
+      var { ipFamily, ipAddress, data } = protocolDecoder(buf)
```
mcollina: This needs to be done only for the frist packet and it should be put behind an option.

test/connect.js
```diff
@@ -416,3 +419,245 @@ test('websocket clients have access to the request object', function (t) {
...
+      // host: parsedProto.proxyAddress.destinationAddress.address.join('.),
+      timeout: 250
+    }
+  )
+
+  client.on('timeout', function () {
+    client.write(protocol)
+  })
+
+  var timer = setTimeout(finish, 1000)
```
mcollina: Can you rewrite this to avoid setTimeout? it makes the tests more brittle.

test/connect.js
```diff
@@ -416,3 +419,245 @@ test('websocket clients have access to the request object', function (t) {
...
+      finish()
+    } else {
+      t.fail('no ip address present')
+    }
+  })
+
+  var client = net.createConnection(
+    {
+      port,
+      // host: parsedProto.proxyAddress.destinationAddress.address.join('.),
```
mcollina: why is this commented out?

lib/client.js
```diff
@@ -45,8 +46,89 @@ function Client (broker, conn, req) {
...
+        that.isProxied = false
+        proto.ipAddress = socket._socket.remoteAddress
+        ipFamily = socket._socket.remoteFamily
+        // console.log('WEBSOCKET ADDRESS', proto.ipAddress)
+      } else if (socket.address) {
+        proto.ipAddress = socket.remoteAddress
+        ipFamily = socket.remoteFamily
+        that.isWebsocket = false
+        that.isProxied = false
+        // console.log('TCPSOCKET ADDRESS', proto.ipAddress)
```
mcollina: remove spurious console.log

lib/client.js
```diff
@@ -45,8 +46,89 @@ function Client (broker, conn, req) {
...
   this.disconnected = false
 
+  this.ipAddress = null
+  this.ipFamily = null
+  this.isWebsocket = false
+  this.isProxied = false
+
   this.parser.on('packet', enqueue)
 
+  function protocolDecoder (data) {
```
mcollina: Can you please move this function out of the constructor? It will cost more memory to execute.

lib/client.js
```diff
@@ -45,8 +46,89 @@ function Client (broker, conn, req) {
 
   this.disconnected = false
 
+  this.ipAddress = null
+  this.ipFamily = null
+  this.isWebsocket = false
+  this.isProxied = false
```
mcollina: Can you put all of these behind a property? Adding more fields to the object will increase our memory footprint per connection

lib/client.js
getlarge: I agree, could you tell me how to detect when the first packet arrived ?
Since _parsingBatch is already decremented, should i change the if (_parsingBatch <= 0) ? 
How do you see this option, like preConnect, Authenticate, etc ... broker hooks ? Or if broker.trustProxy is true ?

test/connect.js
getlarge: Just for the setTimeout, not the net.socket timeout ?

test/connect.js
getlarge: It was intended to show how to manipulate the object. It's now into the example so i'll remove it.

examples/proxy/index.js
getlarge: done

lib/client.js
getlarge: done

lib/client.js
getlarge: An already existing one, or something like this.connDetails : {} ?

lib/client.js
getlarge: It is now in an external module -> protocol-decoder.js.

lib/protocol-decoder.js
```diff
@@ -0,0 +1,84 @@
...
+    client.connDetails.isProxied = true
+  } else if (trustProxy) {
+    var proxyProto
+    try {
+      proxyProto = proxyProtocol.V1BinaryProxyProtocol.parse(data)
+    } catch (V1BinaryProxyProtocolE) {
+      try {
+        proxyProto = proxyProtocol.V2ProxyProtocol.parse(data)
+      } catch (V2ProxyProtocolE) {
+        // empty;
```
mcollina: Why is this empty?

lib/protocol-decoder.js
```diff
@@ -0,0 +1,84 @@
+'use strict'
+
+var proxyProtocol = require('proxy-protocol-js')
+
+function protocolDecoder (client, data) {
+  var proto = {}
+  // var buffer = Buffer.allocUnsafe(0)
+  // buffer = client.conn.read(null)
+  if (!data) return proto
+  // todo: checkProxiesList(client.conn, client.broker.trustedProxies)
```
mcollina: What is this todo?

lib/protocol-decoder.js
```diff
@@ -0,0 +1,84 @@
+'use strict'
+
+var proxyProtocol = require('proxy-protocol-js')
+
+function protocolDecoder (client, data) {
+  var proto = {}
+  // var buffer = Buffer.allocUnsafe(0)
+  // buffer = client.conn.read(null)
```
mcollina: Can you remove these?

lib/protocol-decoder.js
getlarge: V2ProxyProtocol.parse throws error in case of an invalid buffer and the "empty" is due to an habit with my lint rules, forbidding empty blocks.
But not in this project apparently, so this comment will be removed.

mcollina: I'm sorry, I don't understand. Will this error being thrown during normal operations?

lib/protocol-decoder.js
getlarge: Just a reminder to talk about another feature, this one would provide another hook to check server ( proxy ) address and compare it with a proxy whitelists / blacklists ( if the list and the callback are defined ).

lib/protocol-decoder.js
getlarge: Of course

lib/protocol-decoder.js
mcollina: This behavior will slow down aedes by at least 50% - we need a solution that does not generate an exception. The best case would be to feed data into the proxy decoder until it can decode correctly, and then avoiding it completely.

test/connect.js
mcollina: just for the setTimeout.

lib/protocol-decoder.js
getlarge: Allright, i admit it was the lazy approach ;)
Another one could be to check the packet signature and if valid aedes will pick the right parser, ( since the error is thrown by proxy-protocol.js library when the signature is invalid )

mcollina:
Good work!

A couple of things:

1. CI is failing, can you take a look?
2. If the proxy packet is split into two separate chunks, would all this logic work?

getlarge:
Concerning the CI, i noticed that test failures happened with node v12.13.1 and the same happens locally on my computer and they also appear on the master branch.
For now i don't see where to start debugging, but i will dig into it.

For the split, you think of one chunk containing the proxy protocol and the other containing the mqtt packet ? Before the parsing ?
Currently, this is what happened after proxyProtocol.parse, where the "clean" tcp packet is extracted in the data property of the result, which can be sent to mqttPacket.parser.

mcollina:
I mean if the proxy protocol header is split into 2 chunks. Are there some guarantees this cannot happen?

getlarge:
Quote from : https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt
"Both [ proxy protocol v1 and v2] formats are designed to fit in the smallest TCP segment that any TCP/IP host is required to support (576 - 40 = 536 bytes). This ensures that the whole
header will always be delivered at once when the socket buffers are still empty
at the beginning of a connection".

Still it might depend on the proxy configuration i suppose.

There are no guarantee that this cannot happen, so the connection to aedes should be dropped or delayed until the proxy header is complete ?

mcollina:
Given on that text, I think this is ok and I was worrying for nothing.

test/basic.js
```diff
@@ -157,13 +157,17 @@ test('unsubscribe without subscribe', function (t) {
...
-  var s = noError(connect(setup(), opts), t)
+  // var s = noError(connect(setup(), opts), t)
+  var s = connect(setup(), opts)
 
   subscribe(t, s, 'hello', 0, function () {
-    s.conn.emit('close')
+    // s.conn.emit('close')
+    s.conn.destroy(null, function () {
+      t.pass('closed streams')
+    })
```
getlarge: I've been trying to resolve the failing tests and it led me to think that s.conn.emit('close') was not closing streams properly. By replacing this event by the destroy method ( which will also emit the same event after ensuring streams are destroyed ), it seems that no data is received by the writeStream, but it triggered a clientError from the broker which in turns kills the test execution due to noError. So i removed it and the the 2 tests concerned are passing.
Is it an acceptable fix ?

mcollina: this is ok!

examples/proxy/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "getlarge",
+  "license": "MIT",
+  "dependencies": {
+    "aedes": "git+https://git@github.com/getlarge/aedes.git#proxy_and_ip_decoder",
+    "mqemitter": "^3.0.0",
+    "mqtt-packet": "^6.2.1",
+    "proxy-protocol-js": "^4.0.2"
```
mcollina: is this dependency needed here?

lib/client.js
```diff
@@ -45,6 +45,8 @@ function Client (broker, conn, req) {
 
   this.disconnected = false
 
+  this.connDetails = {}
```
mcollina: Can you initialize this to null?

lib/protocol-decoder.js
mcollina: Would you mind removing this todo and opening an issue instead?

test/meta.js
```diff
@@ -26,7 +26,9 @@ test('count connected clients', function (t) {
 
       // needed because destroy() will do the trick before
       // the next tick
-      process.nextTick(function () {
+      // replaced process.nextTick with setImmediate,
+      // before there were still 2 clients connected after the tick, is this fix wrong ?
```
mcollina: Can you remove the part about “replacing nextTick”

test/basic.js
```diff
@@ -157,13 +157,17 @@ test('unsubscribe without subscribe', function (t) {
 })
 
 test('unsubscribe on disconnect for a clean=true client', function (t) {
-  t.plan(6)
+  t.plan(7)
 
   var opts = { clean: true }
-  var s = noError(connect(setup(), opts), t)
+  // var s = noError(connect(setup(), opts), t)
```
mcollina: Can you remove this?

examples/proxy/index.js
```diff
@@ -0,0 +1,180 @@
...
+      timeout: 150
+    }
+  )
+
+  var data = protocol
+  // if (!Buffer.isBuffer(protocol)) {
+  //   data = Buffer.from(protocol.buffer)
+  // } else {
+  //   data = protocol
+  // }
```
mcollina: Can you remove this?

test/basic.js
```diff
@@ -157,13 +157,17 @@ test('unsubscribe without subscribe', function (t) {
...
+  var s = connect(setup(), opts)
 
   subscribe(t, s, 'hello', 0, function () {
-    s.conn.emit('close')
+    // s.conn.emit('close')
```
mcollina: please remove

examples/proxy/package.json
getlarge: It's needed to build proxy protocol packet, as the purpose of this example is to mock what a proxy server sends.

lib/client.js
getlarge: Yes, what's the reason for this ?

test/meta.js
getlarge: done

examples/proxy/index.js
getlarge: done

test/basic.js
getlarge: done

PR #337, opened 2019-12-02
Unauthenticated connection will unregister the current connected one

seeLuck:
This commit fix bug:

Two connection with same client id. Connection A passed `authenticate` and connect to broker successfully. Connection B reject by `authenticate` and failed to connect to broker.

While with this failed Connection B, the current client info will be unregisted in broker.

Few lines to fix it.


seeLuck:
The CI failed with `not ok 333 one connected clients`.

But it is not my case. The master branch also failed with this one.

mcollina:
Would you mind adding a unit test?

--- We'll need a fix for the failing test as well.

seeLuck:
done :)
@mcollina

robertsLando:
@seeLuck there is one test that is failing

seeLuck:
@robertsLando 
It seems not mine problem, I just merge the branch and all passed.

robertsLando:
There was some problem with tests that have been fixed by a recent PR, now should be good

seeLuck:
@mcollina 
Hi boss, how about merge this bugfix commit and update npm package? Thx a lot.

mcollina:
Landed, I hope I can get a new release out soon.

Issue #331, opened 2019-10-23
Wrong clients status


robertsLando:
Hi, I'm using Aedes broker on my project https://github.com/robertsLando/Mqtt2Mqtt

Some mqtt devices that I use doesn't support will message so I have set up a rule on the broker that once a client disconnects/connect sends a message in a specific topic with the clientID with true/false based on client connection status. 

The problem is that sometimes it happens that I receive a disconnect from a client but the client is still connected as he sends/receives packets correctly. Is this a known issue? 

Client connect: https://github.com/robertsLando/Mqtt2Mqtt/blob/master/lib/broker.js#L38
Client disconnect: https://github.com/robertsLando/Mqtt2Mqtt/blob/master/lib/broker.js#L52

robertsLando:
@mcollina I'm still expecting some strange behaviour with client connection management, could you check this issue more in deep? 

I will try to simulate it someway but I can't reproduce it successfully for now.

mcollina:
Unfortunately I have no time to check this. I'm happy to review a PR, if you happen to fix it.

robertsLando:
@gnought Could you help with this? I'm creating a test to replicate the problem.

robertsLando:
@mcollina Here the example code I have created: 


```js
var mqtt = require('mqtt')
var aedes = require('aedes')()
var server = require('net').createServer(aedes.handle)
var Promise = require('bluebird')
var port = 1883

var URL = 'mqtt://localhost';
var CLIENTS = {}

var totalClients = 100;
var totalConnections = 50;
var totalPublish = 30;

// fired when a client connects
aedes.on('client', function (client) {
    var cId = client ? client.id : null;
    console.log('Client Connected: \x1b[33m' + cId + '\x1b[0m')

    if (CLIENTS.hasOwnProperty(cId)) {
        CLIENTS[cId].cCount++
    }
})

// fired when a client disconnects
aedes.on('clientDisconnect', function (client) {
    var cId = client ? client.id : null;
    console.log('Client Disconnected: \x1b[33m' + cId + '\x1b[0m')

    if (CLIENTS.hasOwnProperty(cId)) {
        CLIENTS[cId].dCount++
    }
})

// aedes.on('subscribe', function (subscriptions, client) {
//     console.log('MQTT client \x1b[32m' + (client ? client.id : client) +
//         '\x1b[0m subscribed to topics: ' + subscriptions.map(s => s.topic).join('\n'))
// })

[...]

PR #351, opened 2020-01-14
Delete existing will message on client connect

robertsLando:
#331

### DIFF
--- a/lib/handlers/connect.js
+++ b/lib/handlers/connect.js
@@ -186,17 +186,17 @@ function restoreSubs (arg, done) {
 function storeWill (arg, done) {
   var client = this.client
   client.will = client._will
-  if (client.will) {
-    // delete any existing will messages from persistence
-    client.broker.persistence.delWill(client, function () {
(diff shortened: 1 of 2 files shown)


robertsLando:
I also would suggest to keep just 1 retained message in db foreach client. Actually mondodb allows multiple will messages to be stored for the same clientid, In putWill function there should be a findAndUpdate with upsert true

lib/handlers/connect.js
```diff
@@ -187,13 +187,16 @@ function storeWill (arg, done) {
   var client = this.client
   client.will = client._will
   if (client.will) {
-    client.broker.persistence.putWill(
-      client,
-      client.will,
-      done)
-    return
+    // delete any existing will messages from persistence
+    client.broker.persistence.delWill(client, function () {
```
mcollina: I think this should be deleted nevertheless.

robertsLando: Sure

robertsLando:
@mcollina done

PR #352, opened 2020-01-15
Improved will test

robertsLando:


mcollina:
Why this change?

robertsLando:
@mcollina Because the prev version of the test didn't wait for first client disconnect event and broker was auto-closing after 200ms timeout. This test is cleaner because it waits first client disconnect than connects another makes the check and closes the broker after all checks

Issue #357, opened 2020-01-22
[feat] Extend Aedes tests against all mqemitter/persistences


robertsLando:
Use Github Actions to trigger more extended tests for supported aedes-persistences/mqemitters

srmarjani:
Hi
This is not clear enough.
Did you mean , add tests for all of aedes-persistences/mqemitters like ones for redis and mongo?

robertsLando:
Nope. From what I have understand I think @mcollina means that would like that aedes tests need to run using all difference persistences/mqemitters avbailable using [Github actions](https://github.com/features/actions) instead of redis, they allows to create multiple tasks, I think we could handle this by simply add 2 env vars that specify the mqemitter instance and the persistence and edit aedes current test in this way:

In [test/helper.js](https://github.com/moscajs/aedes/blob/master/test/helper.js) 

```
var persistence = require(process.env.AEDES_PERSISTENCE)
var mqemitter = require(process.env.MQEMITTER)
```

Then [here](https://github.com/moscajs/aedes/blob/master/test/helper.js#L16) add the required persistences to aedes options

mcollina:
I would prefer a very different approach.

Having some tests that use mqtt.js to connect to Aedes and very some of the basic functionality. 
Most of the tests here are unit and they expect things to run single-process.

robertsLando:
Ignore my PR so. WIll wait you to start a different repo for them. I think the actions.yml code can be taken as a base for actions (or not) ?

mcollina:
yes, definitely!

Il giorno gio 23 gen 2020 alle ore 10:30 Daniel Lando <
notifications@github.com> ha scritto:

robertsLando:
@mcollina What about `aedes-test` repo? Would you like me to create it?

I also need to understand which tests you would like to add there

mcollina:
Integration/black box test.

Start an _external_ aedes broker and use MQTT.js to validate it works.

robertsLando:
https://github.com/moscajs/aedes-tests

PR #361, opened 2020-01-23
[WIP] Github actions for cross persistence/mqemitter tests #357

robertsLando:


robertsLando:
Now need to replace all `aedes()` in test with an helper method that adds persistences/mqemitters

mcollina:
Most of the tests are unit. We need a new batch of tests that works treating aedes as a black box.

IMHO, I'd place them in another repo.

robertsLando:
@mcollina Ok does it means a complete rewrite of tests?

mcollina:
Not a rewrite, _new_ tests.

robertsLando:
Yes I used the wrong word but that is what I meant

robertsLando:
@mcollina Would you mind consider a part of this PR? I could remove the actions.yml and the process.env vars and keep the centralized aedes instance creation using helper method. This may be useful in future?

mcollina:
I don't think so. The current tests are too tied and this adds unnecessary complexity.

Issue #348, opened 2020-01-01
[bug] Memory leak on client.js 'connect' event


srmarjani:
Hi
I am using [mqtt](https://www.npmjs.com/package/mqtt) as mqtt client.
In my use case I need to send about 1000 QOS1 packet every 4 second.
every 4 seconds I open a new client by [mqtt.connect](https://www.npmjs.com/package/mqtt#connect) then I publish 1000 packets by [p-map](https://www.npmjs.com/package/p-map) by concurrency 100 and at the end I end connection 

But I am encountering below warning at aedes side

```
(node:15528) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 connected listeners added. Use emitter.setMaxListeners() to increase limit
```
And the stack trace is 
```
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 connected listeners added. Use emitter.setMaxListeners() to increase limit
    at _addListener (events.js:247:17)
    at Client.addListener (events.js:263:10)
    at Parser.enqueue (***\node_modules\aedes\lib\client.js:313:17)
    at Parser.emit (events.js:189:13)
    at Parser._newPacket (***\node_modules\mqtt-packet\parser.js:631:10)
    at Parser.parse (***\node_modules\mqtt-packet\parser.js:39:41)
    at Socket.nextBatch (***\node_modules\aedes\lib\client.js:68:23)
    at Socket.emit (events.js:189:13)
    at emitReadable_ (_stream_readable.js:535:12)
    at process._tickCallback (internal/process/next_tick.js:63:19)
```
I think the problem is in this function
```javascript
function enqueue (packet) {
  this.client._parsingBatch++
  // already connected or it's the first packet
  if (this.client.connackSent || this.client._parsingBatch === 1) {
    handle(this.client, packet, this.client._nextBatch)
  } else {
    this.client.on('connected', () => {
      handle(this.client, packet, this.client._nextBatch)
    })
  }
}
```
Is there any problem in opening and closing the connection to broker?

Thanks

robertsLando:
@srmarjani Try to set `concurrency` option of aedes to 200

robertsLando:
@srmarjani Did you fixed? Can this be closed?

srmarjani:
Hi 
I will try to reproduce the bug. because I choose another approach.
But I guess increasing the concurrency is not a solution.
Because concurrency is 100 by default but as you see in warning 11 event on `connected` event of client make problem.

[...]

PR #362, opened 2020-01-23
fix: Memory leak on client.js 'connected' event #348

robertsLando:


lib/client.js
```diff
@@ -48,6 +48,15 @@ function Client (broker, conn, req) {
...
   this.connDetails = null
 
+  // queue packets received before client fires 'connect' event. Prevents memory leaks on 'connect' event
+  this.parser._queue = []
+
+  this.on('connected', () => {
+    while (this.parser._queue.length > 0) {
+      handle(this, this.parser._queue.shift(), this._nextBatch)
+    }
+  })
```
mcollina: Can you please use a top level function for this listener instead of a closure?

lib/client.js
```diff
@@ -48,6 +48,15 @@ function Client (broker, conn, req) {
 
   this.connDetails = null
 
+  // queue packets received before client fires 'connect' event. Prevents memory leaks on 'connect' event
+  this.parser._queue = []
+
+  this.on('connected', () => {
```
mcollina:
```suggestion
  this.once('connected', () => {
```

lib/client.js
```diff
@@ -48,6 +48,15 @@ function Client (broker, conn, req) {
...
+    while (this.parser._queue.length > 0) {
+      handle(this, this.parser._queue.shift(), this._nextBatch)
+    }
```
mcollina: Instead of calling `shift()` just use a `for(;;)` loop and then assign `this.parser._queue` to null.

lib/client.js
```diff
@@ -48,6 +48,20 @@ function Client (broker, conn, req) {
...
+  this.on('connected', dequeue)
+
+  function dequeue () {
+    var q = this.parser._queue
+    for (var i = 0, len = q.length; i < len; i++) {
+      handle(this, q[i], this._nextBatch)
+    }
+
+    this.parser._queue = null
+  }
```
mcollina: Please make this top-level, outside of this function.

lib/client.js
```diff
@@ -48,6 +48,20 @@ function Client (broker, conn, req) {
 
   this.connDetails = null
 
+  // queue packets received before client fires 'connect' event. Prevents memory leaks on 'connect' event
+  this.parser._queue = []
+
+  this.on('connected', dequeue)
```
mcollina:
```suggestion
  this.once('connected', dequeue)
```

lib/client.js
robertsLando: You mean the entire function dequeue?

mcollina:
The code is ok, the test on the EventEmitter warning is missing

robertsLando:
@mcollina What do you mean? 

This ?

```
process.on('warning', t.fail);
```

robertsLando:
Without the queue fix the test fails so it tests also EventEmitter warning now

robertsLando:
@mcollina Can I merge this or need to wait reviwers?

mcollina:
please wait for a bit

lib/client.js
```diff
@@ -48,6 +48,11 @@ function Client (broker, conn, req) {
 
   this.connDetails = null
```
GavinDmello: this.connDetails can be removed if not used.

gnought:
Why we need to queue all packets between CONNECT and connected event?
We emit the "connected" event at https://github.com/moscajs/aedes/blob/f62a34324bbefd6060eb52a8489d5c77e19e6eaf/lib/handlers/connect.js#L93. After we receive CONNECT and trigger connect handler, we still have do authentication, what will happen for the queue packets if the authentication fails?

lib/client.js
robertsLando: It is used [here](https://github.com/moscajs/aedes/blob/master/lib/protocol-decoder.js#L116)

And @mcollina suggested @getlarge to init it to null in client in his PR

robertsLando:
@gnought `connected` event is trigger after [connectActions](https://github.com/moscajs/aedes/blob/master/lib/handlers/connect.js#L22) 

We need to queue packets because like you can see in issue #348 after the first connect packet is received all other packet init a listener to connect event [here](https://github.com/moscajs/aedes/blob/master/lib/client.js#L333) and this throws the eventemitter warning

gnought:
The memory leakage occurs obviously we are keeping add listener in a same 'connected' event, and over the default limit 10 in https://nodejs.org/docs/latest-v12.x/api/events.html#events_emitter_setmaxlisteners_n

Currently we hook `packet` event on https://github.com/moscajs/aedes/blob/master/lib/client.js#L51, we parse all pending packets once 'connected' is triggered.

I think we should drop all packets before there is a `clientReady` or `client` event occur
That means in https://github.com/moscajs/aedes/blob/master/lib/client.js#L333-L335, we should change to 
```
    if (client.connected)
      handle(client, packet, client._nextBatch)
```

robertsLando:
@gnought @mcollina Dropping packets could be an option but I don't know if this is spec compilant. Should we reject packets before the client is connected successfully? Should client wait to receive the connack before sending other commands? If so we can drop packets

gnought:
Rejecting packets before "connected" event makes sense. We could get rid of unauthenticated packets. Clients should see if the connection is ready like ioredis ready event.
Users may make use of `preConnect` to implement their protection logic, otherwise we need to think a way how to clear those queued packets if it is unauthenticated client.

robertsLando:
@gnought Wait but in case on unauthenticated client the connection is closed so where is the problem with the cache?

GavinDmello:
> The memory leakage occurs obviously we are keeping add listener in a same 'connected' event, and over the default limit 10 in https://nodejs.org/docs/latest-v12.x/api/events.html#events_emitter_setmaxlisteners_n
> 
> Currently we hook `packet` event on https://github.com/moscajs/aedes/blob/master/lib/client.js#L51, we parse all pending packets once 'connected' is triggered.
> 
> I think we should drop all packets before there is a `clientReady` or `client` event occur
> That means in https://github.com/moscajs/aedes/blob/master/lib/client.js#L333-L335, we should change to
> 
> ```
>     if (client.connected)
>       handle(client, packet, client._nextBatch)
> ```

Was going to say, this would be dangerous as we would be parsing packets from unauthenticated clients. 
We must ensure that the queue is being flushed if there are connection/client errors.

gnought:
@GavinDmello Thanks for pointing out.

@robertsLando Try this.
```
test.only('closed client - queued items are left behind', function (t) {
  var evt = new EE()
  var broker = aedes({
    authenticate: async (client, username, password, done) => {
      await sleep(2000) // simulate network
      done(new Error('Access Denied'), false)
      evt.emit('AuthenticateEnd', client)
    }
  })

  var s = connect(setup(broker, false))
  var publishP = {
    cmd: 'publish',
    topic: 'hello',
    payload: Buffer.from('world'),
    qos: 0,
    retain: false
  }
  for (let i = 0; i < 100; i++) {
    s.inStream.write(publishP)
  }

  evt.on('AuthenticateEnd', function (client) {
    t.equal(broker.connectedClients, 0)
    t.equal(client.parser._queue.length, 0)
    setImmediate(() => {
      broker.close()
    })
  })
  broker.on('closed', () => {
    t.end()
  })
})

**Comment by robertsLando** [comment]
That is right, but I can free it on client close event

**Comment by gnought** [comment]
Yes, we can, but the fix does not convince me. 
If it is under attack and there are lots of delays in preConnect per client with a high volume traffic. All queues are growing bigger and consumes a lot of memory, and we will let  Node to use extra power to cleanup all items, even there may be a message lag if Node does not process faster enough.

**Comment by mcollina** [comment]
I think we can error the client after `42` messages being queued, making it configurable by the user.

**Comment by GavinDmello** [comment]
> I think we can error the client after `42` messages being queued, making it configurable by the user.

@mcollina Life, universe and everything ?

I'm 👍 for configurable amount of messages.

**Comment by robertsLando** [comment]
@mcollina Should be ok now?

**Comment by gnought** [comment]
It sounds like that we still queue messages after connected.
What happen when the incoming messages are faster than the processing rate?

**Comment by robertsLando** [comment]
> It sounds like that we still queue messages after connected.

After connected event all queued packets are processed and the queue is set to null

> What happen when the incoming messages are faster than the processing rate?

I think that this problem would be the same with or without the initial queue, if the incoming messages are faster then the processing rate simply we get a stack overflow but this would not be caused by the queue

**Comment by robertsLando** [comment]
@mcollina So what about this?

File: lib/client.js [hunk 7]
```diff
@@ -48,6 +48,12 @@ function Client (broker, conn, req) {
...
 
+  // queue packets received before client fires 'connect' event. Prevents memory leaks on 'connect' event
+  this.parser._queue = []
+  this.parser.queueLimit = broker.queueLimit
```
mcollina: There is no need to copy this, we have `this.broker` where we can access this value from. I prefer to not add too many fields here because it will increase the memory footprint for every client.

lib/client.js
```diff
@@ -330,9 +338,21 @@ function enqueue (packet) {
   if (client.connackSent || client._parsingBatch === 1) {
     handle(client, packet, client._nextBatch)
   } else {
-    client.on('connected', () => {
-      handle(client, packet, client._nextBatch)
-    })
+    if (this._queue.length < this.queueLimit) {
```
mcollina:
```suggestion
    if (this._queue.length < client.broker.queueLimit) {
```

lib/client.js
```diff
@@ -330,9 +338,21 @@ function enqueue (packet) {
...
+      this.emit('error', new Error('Client queue limit reached'))
+    }
+  }
+}
+
+function dequeue () {
+  if (this.parser._queue) {
+    for (var i = 0, len = this.parser._queue.length; i < len; i++) {
+      handle(this, this.parser._queue[i], this._nextBatch)
+    }
```
mcollina: Can you please move `this.parser._queue` to an variable?

robertsLando:
@mcollina Why? I'm using standardjs formatter for my workspace and it does that change

robertsLando:
@mcollina done

robertsLando:
@mcollina I don't know which ide you are using, if everyone is using vscode we could add .vscode folder with all the formatters used (maybe another PR for this)

mcollina:
I prefer integrations that are validated by CI, not by the editor.

mcollina:
(I use vim)

mcollina:
this conflicts with master now

robertsLando:
@mcollina It doesn't really conflicts it just fails the coverall tests as now `lib/handlers/connect.js` has reduced coverage https://coveralls.io/builds/28372782

robertsLando:
The coverage is decreesed by 0.04% causing that red cross on ci bu all other tests are good. Should be create a separate job for coverage?

mcollina:
increase unit test coverage
Il 29 gen 2020, 08:10 +0100, Daniel Lando <notifications@github.com>, ha scritto:

PR #366, opened 2020-01-24
[chore] Prevent double client close call

gnought:


lib/client.js
```diff
@@ -232,8 +232,10 @@ Client.prototype.unsubscribe = function (packet, done) {
 
 Client.prototype.close = function (done) {
   var that = this
-  var conn = this.conn
 
+  if (this._eos === nop) { return }
```
mcollina: This branch must call the done function

PR #369, opened 2020-01-25
[chore] Reject 2nd CONENCT while in preConnect

gnought:
We have already followed [MQTT-3.1.0-2] spec to block 2nd CONNECT. This only applies to when `connected=true` clients. This PR is to stop any 2nd CONNECT even in the early preConnect stage where `connected=false`

### DIFF
--- a/lib/handlers/index.js
+++ b/lib/handlers/index.js
@@ -11,18 +11,18 @@ var handlePing = require('./ping')
 
 function handle (client, packet, done) {
   if (packet.cmd === 'connect') {
-    if (client.connected) {
+    if (client.connected || !client._connectTimer) {
       // [MQTT-3.1.0-2]
-      client.conn.destroy()
-      return done(new Error('Invalid protocol'))
+      finish(client.conn, packet, done)
+      return
     }
     handleConnect(client, packet, done)
     return
   }
   if (!client.connected) {
     // [MQTT-3.1.0-1]
-    client.conn.destroy()
-    return done(new Error('Invalid protocol'))
+    finish(client.conn, packet, done)
+    return
   }
 
   switch (packet.cmd) {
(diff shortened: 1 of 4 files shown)


test/helper.js
```diff
@@ -114,10 +114,13 @@ function subscribeMultiple (t, subscriber, subs, expectedGranted, done) {
   })
 }
 
+const delay = (time) => new Promise((resolve) => setTimeout(resolve, time))
```
mcollina: Can we please use util.promisify instead?

test/connect.js
```diff
@@ -307,6 +308,53 @@ test('second CONNECT Packet sent from a Client as a protocol violation and disco
...
+  broker.on('connectionError', function (client, err) {
+    t.equal(err.info.clientId, 'my-client-2')
+    t.equal(err.message, 'Invalid protocol')
+  })
+
+  const msg = (s, ms, msg) => {
+    return new Promise((resolve) => {
+      setTimeout(() => { s.inStream.write(msg); resolve() }, ms)
+    })
+  }
```
mcollina: please use the delay utility and async/await

test/connect.js
```diff
@@ -307,6 +308,53 @@ test('second CONNECT Packet sent from a Client as a protocol violation and disco
...
+
+  Promise.all([msg(s, 100, packet1), msg(s, 200, packet2)]).then(() => {
+    broker.close()
+    t.end()
+  })
```
mcollina: please add a catch() to this promise.all.

lib/handlers/index.js
```diff
@@ -62,6 +62,13 @@ function handle (client, packet, done) {
   if (client._keepaliveInterval > 0) {
     client._keepaliveTimer.reschedule(client._keepaliveInterval)
   }
+
+  function finish (conn, packet, done) {
+    conn.destroy()
+    var error = new Error('Invalid protocol')
+    error.info = packet
+    done(error)
+  }
```
mcollina: please avoid creating a closure here, they use memory and this is a hot path.

gnought: sure, I haven't noticed that. :P

PR #375, opened 2020-01-26
[chore] Mark the time before we create aedes in keepalive test

gnought:


gnought:
The test sometimes gets fail, even in my local machine. I think we start the time and do the rest will be avoid some races.

PR #379, opened 2020-01-30
[chore] Stop series of connect actions asap if broker is closed

gnought:
Once broker is closed while in authentication, we stop processing pending actions (e.g. set keepalive, restore subscriptions, register clients etc.) as quick as possible in resource wise.

### DIFF
--- a/test/close_socket_by_other_party.js
+++ b/test/close_socket_by_other_party.js
@@ -7,15 +7,15 @@ var aedes = require('../')
 var setup = helper.setup
 var connect = helper.connect
 var subscribe = helper.subscribe
-var delay = helper.delay
 
 test('aedes is closed before client authenticate returns', function (t) {
   t.plan(0)
 
   var broker = aedes({
-    authenticate: async (client, username, password, done) => {
-      await delay(2000) // simulate network
-      done(null, true)
+    authenticate: (client, username, password, done) => {
+      setTimeout(function () {
+        done(null, true)
+      }, 2000)
     }
   })
   broker.on('client', function (client) {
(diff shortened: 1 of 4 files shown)


test/close_socket_by_other_party.js
```diff
@@ -9,6 +9,31 @@ var connect = helper.connect
...
 var delay = helper.delay
 
+test('aedes is closed before client authenticate returns', function (t) {
+  t.plan(0)
+
+  var broker = aedes({
+    authenticate: async (client, username, password, done) => {
+      await delay(2000) // simulate network
+      done(null, true)
+    }
```
mcollina: Please do not use an async function. It's a bad pattern because there will be no one checking for an error message, resulting in an unhandledRejection. CHeck out https://github.com/mcollina/make-promises-safe for more details.

test/will.js
```diff
@@ -277,29 +298,32 @@ test('does not deliver a will without authentication', function (t) {
...
   var opts = {}
   opts.keepalive = 1
   var broker = aedes({
-    authenticate: function (c, u, p, cb) {
-      setTimeout(function () {
-        cb(null, true)
-      }, 3000)
+    authenticate: async function (client, username, password, callback) {
+      await delay(3000)
+      callback(null, true)
```
mcollina: do not use async/await.

test/close_socket_by_other_party.js
gnought: Does it mean that we also should not use async at here https://github.com/moscajs/aedes/blob/master/test/connect.js#L333 ?

mcollina: exactly. None of the aedes-codebase if promise friendly. We should possibly add support for those in the future.

gnought: sounds like we need to replace the delay with setTimeout in this stage.
Which ways do you prefer to make aedes-codebase be promis friendly?

mcollina: open a new issue to discuss!

gnought:
@mcollina does it look good to you after dropping the async/await pattern?

Issue #338, opened 2019-12-04
Validate proxies list before protocol decoder


getlarge:
Would you mind removing this todo and opening an issue instead?

_Originally posted by @mcollina in https://github.com/mcollina/aedes/pull/334_

robertsLando:
@getlarge Would you mind to submit a PR for this?

getlarge:
Yes i might make a proposal next week.
By the way, there is still an issue in the with the protocol decoder when Aedes broker is used behind a TCP proxy which handles SSL termination.
The packet is splitted in two, on one side the protocol proxy header and the TCP buffer on the other.
Any help to solve this would be appreciated.

robertsLando:
Would you mind to open a new issue about that? Just to keep track of it

Daniel

robertsLando:
@getlarge should this be moved to aedes-server-factory or protocol decoder libs instead?

getlarge:
Good question, protocol-decoder might be a better place. We could still reuse that logic inside the server-factory via an option that could be overridden by user like `protocolDecoder`, `serverFactory`.

PR #380, opened 2020-01-30
Fix protocolDecoder phase

### LINKED ISSUE #364: Protocol decoder not always working with TCP proxy
There is still an issue in the the protocol decoder when Aedes broker is used behind a TCP proxy which handles SSL termination.
The packet is splitted in two, on one side the protocol proxy header and the TCP buffer on the other.
Any help to solve this would be appreciated.

_Originally posted by @getlarge in https://github.com/moscajs/aedes/issues/338#issuecomment-577841457_

getlarge:
- Removed assignement destructuring ( in nextBatch ) in case a developer provide its own protocolDecoder that would return something else than an object.

- Changed packet parsing condition to fix #364 

- Extract serverAddress in protocolDecoder ( use forwarded package for WS headers ) to prepare for #338 

- Added test mocking TCP proxy behaviour, but might require a test with Nginx and/or HAproxy

### DIFF
--- a/lib/protocol-decoder.js
+++ b/lib/protocol-decoder.js
@@ -1,6 +1,7 @@
 'use strict'
 
 var proxyProtocol = require('proxy-protocol-js')
+var forwarded = require('forwarded')
 
 var v1ProxyProtocolSignature = Buffer.from('PROXY ', 'utf8')
 var v2ProxyProtocolSignature = Buffer.from([
@@ -59,9 +60,16 @@ function protocolDecoder (client, data) {
     var headers = client.req && client.req.headers ? client.req.headers : null
     var proxyProto
     if (headers) {
-      if (headers['x-real-ip']) proto.ipAddress = headers['x-real-ip']
-      else if (headers['x-forwarded-for']) proto.ipAddress = headers['x-forwarded-for']
+      if (headers['x-forwarded-for']) {
+        var addresses = forwarded(client.req)
+        proto.ipAddress = headers['x-real-ip'] ? headers['x-real-ip'] : addresses[addresses.length - 1]
+        proto.serverIpAddress = addresses[0]
+      }
+      if (headers['x-real-ip']) {
+        proto.ipAddress = headers['x-real-ip']
+      }
       proto.port = socket._socket.remotePort
+      ipFamily = socket._socket.remoteFamily
       proto.isWebsocket = true
     }
     if (isValidV1ProxyProtocol(data)) {
@@ -70,6 +78,7 @@ function protocolDecoder (client, data) {
         ipFamily = proxyProto.inetProtocol
         proto.ipAddress = proxyProto.source.ipAddress
         proto.port = proxyProto.source.port
+        proto.serverIpAddress = proxyProto.destination.ipAddress
         proto.data = proxyProto.data
         proto.isProxy = 1
       }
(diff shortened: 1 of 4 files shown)


test/connect.js
```diff
@@ -897,3 +898,101 @@ test('websocket proxied clients have access to the ipAddress from x-forwarded-fo
...
+  }, function () {
+    client.write(buf)
+  })
+
+  client.on('timeout', function () {
+    client.end(mqttPacket.generate({ cmd: 'disconnect' }))
+  })
+
+  function finish () {
+    // client.end(mqttPacket.generate({ cmd: 'disconnect' }))
```
mcollina: why is this commented?

getlarge: a mistake, i forgot to remove it

robertsLando:
@getlarge about our discussion [here](https://github.com/moscajs/aedes/issues/364#issuecomment-580167785) what about create an option to force using protocol decoder? With previous version it was used only with trustProxy option set to true, now it is used everytime there is a valid packet, it could add some complexity that is not needed by most users

getlarge:
@robertsLando That makes sense, but i'm not sure adding another option is the right way, the list is getting long.
Maybe we could export the protocol decoder ( or make a package for it ), and if someone wants to use it, he/she would provide its own decodeProtocol in the broker option.
Then defaultDecodeProtocol would return null for example.
What do you think of this ?

robertsLando:
Let's check what @mcollina thinks about this. I'm +1 for both solutions, I would like to keep aedes as slim as possible. Protocol decoder could become an aedes plugin for sure

mcollina:
I’m fine either way as long as there is no overhead if it’s not enabled.

getlarge:
Since the part that would handle authorize proxies ( #338 ) might also use external dependencies, it could be logical to include it in this decoder plugin, and keep aedes light.

Currently the protocolDecoder, in aedes master branch, is never used as long broker.trustProxy is false. So the behaviour would be the same by making it external, per default it would be a function that would return null.

If you decide it's the way to go, would you mind creating a boilerplate repo ? 
Aedes-protocol-parser could be a good name ?
Then i will include the protocol-decoder.js and transfer related tests from aedes in it.

robertsLando:
I can create it for you.

robertsLando:
@getlarge  Here you are: https://github.com/moscajs/aedes-protocol-decoder

lib/protocol-decoder.js
```diff
@@ -59,9 +60,16 @@ function protocolDecoder (client, data) {
     var headers = client.req && client.req.headers ? client.req.headers : null
     var proxyProto
     if (headers) {
-      if (headers['x-real-ip']) proto.ipAddress = headers['x-real-ip']
-      else if (headers['x-forwarded-for']) proto.ipAddress = headers['x-forwarded-for']
+      if (headers['x-forwarded-for']) {
+        var addresses = forwarded(client.req)
```
GavinDmello: Could we extract the addresses without the use of the `forwarded` package ?

getlarge: Yes sure, but why ?
As you can see [forwarded](https://github.com/jshttp/forwarded/blob/master/index.js) is a very small module.

robertsLando: In my opinion keep forwarded is ok

robertsLando:
@getlarge @mcollina Could we close this?

getlarge:
@robertsLando I think so, do you want me to push a cleaned version of aedes?
I also added a bit of doc on readme and some types.

robertsLando:
A PR is already ready for that :) just need to publish aedes-protocol-decoder to npm

Daniel

robertsLando:
@getlarge aedes-protocol-decoder has been published and #395 is the aedes PR to support it

PR #385, opened 2020-01-31
Added a proper catch block for handling errors

gnought:
@mcollina Is it better to add .catch() block rather than try/catch?
any differences?

### DIFF
--- a/test/connect.js
+++ b/test/connect.js
@@ -12,6 +12,7 @@ var mqtt = require('mqtt')
 var mqttPacket = require('mqtt-packet')
 var net = require('net')
 var proxyProtocol = require('proxy-protocol-js')
+var util = require('util')
 
 ;[{ ver: 3, id: 'MQIsdp' }, { ver: 4, id: 'MQTT' }].forEach(function (ele) {
   test('connect and connack (minimal)', function (t) {
@@ -348,10 +349,11 @@ test('reject second CONNECT Packet sent while first CONNECT still in preConnect 
     await delay(ms)
     s.inStream.write(msg)
   }
+  const immediate = util.promisify(setImmediate)
 
-  (async () => {
+  ;(async () => {
     await Promise.all([msg(s, 100, packet1), msg(s, 200, packet2)])
-    setImmediate(() => {
+    await immediate().then(() => {
       broker.close()
       t.end()
     })
--- a/test/connect.js
+++ b/test/connect.js
@@ -350,16 +350,16 @@ test('reject second CONNECT Packet sent while first CONNECT still in preConnect 
   }
 
   (async () => {
-    try {
-      await Promise.all([msg(s, 100, packet1), msg(s, 200, packet2)])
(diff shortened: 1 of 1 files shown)


mcollina:
the try catch will save some promise allocation. However there is a risk that the catch part in the try/catch throw, resulting in an unhandled rejection.

This code is fine, it's just for tests.

gnought:
Thanks for the explanation. Could we stick into a practise and merge it into the master branch?

test/connect.js
```diff
@@ -350,16 +350,16 @@ test('reject second CONNECT Packet sent while first CONNECT still in preConnect
...
   (async () => {
-    try {
-      await Promise.all([msg(s, 100, packet1), msg(s, 200, packet2)])
-      setImmediate(() => {
-        broker.close()
-        t.end()
-      })
-    } catch (error) {
+    await Promise.all([msg(s, 100, packet1), msg(s, 200, packet2)])
+    setImmediate(() => {
```
mcollina: I would make this a promise and use await.

gnought:
```
  function finish () {
    return new Promise((resolve, reject) => {
      broker.close()
      resolve(t.end)
    })
  }
  (async () => {
    await Promise.all([msg(s, 100, packet1), msg(s, 200, packet2)])
    await finish()
  })().catch(
    (error) => {
      t.fail(error)
    }
  )
```
Is the above what you mean? :)

gnought: or should it be like this?
```
  (async () => {
    await Promise.all([msg(s, 100, packet1), msg(s, 200, packet2)])
  })().catch(
    (error) => {
      t.fail(error)
    }
  ).finally(finish)
```

mcollina: I mean

```js
const immediate = promisify(setImmediate)
```

gnought: awesome!. I haven't seen we could promisify `setImmediate`. 
Like this?
```
  const immediate = util.promisify(setImmediate)

  ;(async () => {
    await Promise.all([msg(s, 100, packet1), msg(s, 200, packet2)])
    await immediate().then(() => {
      broker.close()
      t.end()
    })
  })().catch(
    (error) => {
      t.fail(error)
    }
  )
```

gnought: > I would make this a promise and use await.

May we know what the differences are?

test/connect.js
```diff
@@ -348,10 +349,11 @@ test('reject second CONNECT Packet sent while first CONNECT still in preConnect
     await delay(ms)
     s.inStream.write(msg)
   }
+  const immediate = util.promisify(setImmediate)
 
-  (async () => {
+  ;(async () => {
     await Promise.all([msg(s, 100, packet1), msg(s, 200, packet2)])
-    setImmediate(() => {
+    await immediate().then(() => {
```
mcollina:
```suggestion
    await immediate()
```

gnought: awesome. You are a great teacher!

PR #388, opened 2020-02-02
Fixed helper connected callback

gnought:
The connected callback is to simulate the `connected` event, it should be called only when `returnCode === 0`


gnought:
The CI will be fine after merging https://github.com/moscajs/aedes/pull/387

PR #395, opened 2020-02-03
aedes-protocol-decoder plugin support

robertsLando:


robertsLando:
@mcollina CI will be broken until we don't release aedes-protocol-decoder to npm

robertsLando:
@mcollina

PR #402, opened 2020-02-05
Added test backpressure published fn

gnought:


test/events.js
```diff
@@ -122,3 +122,43 @@ test('Emit event when broker closed', function (t) {
...
 })
+
+test('Test backpressure aedes published function', { timeout: 500 }, function (t) {
+  t.plan(1)
+
+  const broker = aedes({
+    published: function (packet, client, done) {
+      if (client) {
+        t.pass('published')
+        setTimeout(done, 1000)
```
mcollina: I don't understand how this timeout relates to the timeout in the actual test.

test/events.js
```diff
@@ -122,3 +122,43 @@ test('Emit event when broker closed', function (t) {
...
+  })
+  t.tearDown(() => {
+    client.end()
+    broker.close()
+    server.close()
+  })
+
+  const mqtt = require('mqtt')
+  const server = require('net').createServer(broker.handle)
+  const port = 1883
```
mcollina: please set this to 0 and get the `port` from `server.address().port`

test/events.js
gnought: this `setTimeout` is to stimulate a backpressure on published function. The whole test timeout is set less than 1000 to let the test plan end with 1 passed result.

mcollina: How long will the test last?

gnought: 500

test/events.js
```diff
@@ -123,30 +123,33 @@ test('Emit event when broker closed', function (t) {
...
   const broker = aedes({
     published: function (packet, client, done) {
       if (client) {
         t.pass('published')
-        setTimeout(done, 1000)
+        setTimeout(() => {
+          publisher.end()
+          done()
+        }
+        , 1000)
```
mcollina: can you remove this?

test/events.js
```diff
@@ -122,3 +122,45 @@ test('Emit event when broker closed', function (t) {
...
+
+  const mqtt = require('mqtt')
+  const server = require('net').createServer(broker.handle)
+  var publisher
+
+  server.listen(0, function () {
+    const port = server.address().port
+    publisher = mqtt.connect({ port: port, host: 'localhost', clean: true, keepalive: 30 })
+
+    var count = 10
```
mcollina: We should verify that count reach a certain value when the test ends, maybe 0?

PR #404, opened 2020-02-05
Dropped pump pkg, use stream pipeline

gnought:
pump has been integrated as stream pipeline since NodeJS v10


lib/handlers/connect.js
```diff
@@ -1,7 +1,7 @@
 'use strict'
 
 const retimer = require('retimer')
-const pump = require('pump')
+const { pipeline } = require('stream')
```
mcollina: we should use the one from the readable-stream module instead (it's already on the dependency tree)

PR #408, opened 2020-02-05
Refactored Subscribe Handler

gnought:


lib/handlers/subscribe.js
```diff
@@ -152,26 +152,24 @@ function completeSubscribe (err) {
...
-
   const subs = packet.subscriptions
 
-  if (!packet.restore) {
-    broker.emit('subscribe', subs, client)
-  }
-
+  // no messageId conditions:
+  // - restore sub, won't have messageId
+  // - programatically trigger subscribe without messageId
```
mcollina: Where does this `restore` property come from? Maybe we could avoid setting it in the first place.

gnought:
I think it is best to use `connecting` flag https://github.com/moscajs/aedes/pull/409 to replace `restore` property

gnought:
Users may hook broker `client` event to do a subscribe. The `connecting` is true during `client` event emits. In this case users cannot subscribe a topic if we use `connecting` to replace `restore` property. As we support offline queueing messages before CONNECT, I abandon this approach.

Previous `restore` is added as an item in packet object in `handleSubscribe` to handle restored subscription behaviour (https://github.com/moscajs/aedes/blob/master/lib/handlers/connect.js#L179), it is not safe if users inject this key. My approach is to remove it, and add one more arguments in `handleSubscribe`. 

In addition `handleSubscribe` is refactored to handle normal and restored subscription separately. Simpler and readable.

gnought:
@mcollina does this PR look goods to you?

PR #413, opened 2020-02-06
WIP: Replace websocket-stream with ws

robertsLando:


robertsLando:
@mcollina I'm getting an error on CI that seems caused by protocol decoder:

`Cannot read property 'remotePort' of undefined`

I see that in `websocket-stream` npm module the socket property is added [here](https://github.com/maxogden/websocket-stream/blob/master/stream.js#L83)

And protocol decoder fails here: 

```js
var socket = conn.socket || conn
....

 proto.port = socket._socket.remotePort
 ipFamily = socket._socket.remoteFamily
```

In poor words how can I get remotePort and remoteFamily props using the ws stream?  I think this should be fixed in protocol decoder so

mcollina:
it should be there somewhere, dig deep

robertsLando:
It should be there but can't find it :confused:  https://nodejs.org/api/net.html#net_socket_remoteaddress

mcollina:
https://github.com/websockets/ws/blob/9535702e50525621313941f0e1a972fdca403045/lib/websocket.js#L150 this should be it.

robertsLando:
@mcollina This way it works but I would much prefer websocket-stream way

mcollina:
The implementation of `ws` is significantly better than the one in websocket-stream. We might want to update the readme with this change as well.

Note: we should ask about adding the socket property to `ws` streams.

robertsLando:
What about create an helper method un aedes to support it? Something like the method used in test/helper? 

Daniel

robertsLando:
@mcollina If you check the issue on ws module I asked to add the property but they told me that they won't do it (check link for more info). Should I also update examples?

mcollina:
I think we should publish an utility module for this purpose. IMHO this is very useful.
Il 7 feb 2020, 16:50 +0100, Daniel Lando <notifications@github.com>, ha scritto:

robertsLando:
You mean “aedes-websockets” plugin? 

It should work like: 

var websocket = require(“aedes-websockets”)

websocket(server, broker)


?


Daniel

robertsLando:
@mcollina So what with this?

mcollina:
I would just create a tiny module that offers the same API of websocket-stream server: https://www.npmjs.com/package/websocket-stream#on-the-server

gnought:
How about we create a factory createServer method to provide websocket, tls function?

robertsLando:
I’m +1 for this and also for the module option, the downside of this is that it would require to add websocket as dependency and some users would not need it. 

Daniel

gnought:
We could build a tiny module to include it

robertsLando:
I can do it tomorrow...

Daniel

gnought:
Maybe we could follow what fastify did. https://github.com/fastify/fastify/blob/master/lib/server.js

gnought:
Haven't done thoughtful check, 

```js
const assert = require('assert')
const WebSocket = require('ws')
const tls = require('tls')
const http = require('http')
const https = require('https')
const net = require('net')

function createServer (options, aedesHandler) {
  assert(options, 'Missing options')
  assert(aedesHandler, 'Missing aedes handler')

  var server = null
  if (options.serverFactory) {
    server = options.serverFactory(aedesHandler, options)
  } else if (options.tls) {
    server = tls.createServer(options, aaedesHandler)
  } else if (options.ws) {
    if (options.https) {
      if (options.http2) {
        server = https.createSecureServer(options.https)
      } else {
        server = https.createServer(options.https)
      }
    } else {
      if (options.http2) {
        server = https.createServer()
        server.on('session', sessionTimeout(options.http2SessionTimeout))
      } else {
        server = http.createServer()
      }
    }
    const ws = new WebSocket.Server({ server: server })
    ws.on('connection', function (conn, req) {
      const stream = WebSocket.createWebSocketStream(conn)
      aedesHandler(stream, req)
    })
  } else {
    server = net.createServer(options, aedesHandler)
  }
  return server
}
```

robertsLando:
@gnought I think that is not necessary to support every protocol, @mcollina was speaking about a module for websockets, that should be enought

robertsLando:
@mcollina Are you ok so for a `aedes-websocket` module?

mcollina:
yes I am

gnought:
As _socket is not available in ws, so there’s a proposal to create a tiny module to provide that function and replace the websocket-stream, but I don’t think it is bound to aedes only. 
Rather I prefer to have a tiny module to provide a createServer function and we offer a serverFactory to let users to provide their own architecture, even including PROXY is nice.

robertsLando:
@gnought Yes but that requires some modules that maybe the user don't need ?

robertsLando:
@mcollina @gnought I would like to work on this in next days but I would like to find something that is good for everone. @mcollina says to create the single websocket module, @gnought prefers a module that will act as server factory for aedes. In my opinion a websocket module is enought as others servers are really easy to setup and that is the only one that requires a little trick

gnought:
How could we handle it in a use case PROXY + websocket ?

robertsLando:
Why? That should be handled in the protocolDecoder module, this is just a replacement for websocket-stream

Daniel

gnought:
For the PROXY+websocket case, likely the order is decoding the PROXY, and process the websocket. The `protocolDecoder` is a helper to decode the PROXY protocol, but it is called in aedes directly, in other words, the order will be websocket and decoding the PROXY if having a aedes-websocket module.

robertsLando:
> For the PROXY+websocket case, likely the order is decoding the PROXY

Why?

The module we are speaking about just creates a websocket stream that is attached to aedes like others streams with `aedes.handle`, so the protocol is decoded inside aedes before the mqtt parser like it works for others protocols like http/https etc..

gnought:
Maybe I'm wrong in the previous comment. The current flow is
1. pass `conn` to aedes (`req` will be available if `conn` is ws), `conn` is a raw stream (could be http/https/tcp/ws)
2. decode the raw stream if PROXY is there.
3. then pass to mqtt parser to read mqtt packets

If we have websocket module, what I understand the flow may become:
1. pass `conn` to aedes (`req` will be available if `conn` is ws), `conn` is a raw stream (could be http/https/tcp/ws)
2. decode the raw stream if PROXY is there.
3. decode websocket if the stream is ws
4. then pass to mqtt parser to read mqtt packets

If it is, I prefer like this.
1. have a connection factory and decode `conn` to be a plain mqtt stream
2. pass `conn` to aedes handle
3. then pass to mqtt parser to read mqtt packets.

robertsLando:
@gnought In poor words you would like the protocol decoding to be done before passing the stream to aedes?

gnought:
The advantage is user could create their own protocol decoder and pipe the decoded stream to aedes.handle. More flexible.
What i posted before, if user wanna websocket over mqtt. Simple.
```
    const ws = new WebSocket.Server({ server: server })
    ws.on('connection', function (conn, req) {
      const stream = WebSocket.createWebSocketStream(conn)
      aedes.handle(stream, req)
    })
```

robertsLando:
I would like to know what @mcollina thinks about this

mcollina:
I don’t understand the problem. Can you frame a clear question? From what I understand the proxy protocol decoding must happen before the websocket even exists.

robertsLando:
> From what I understand the proxy protocol decoding must happen before the websocket even exists

I don't understand what you mean @mcollina. How do you think we should implement protocol decoder so? As an aedes plugin (like it is working now) or like @gnought suggests?

As @gnought said now the flow is:

1. pass `conn` to aedes (`req` will be available if `conn` is ws), `conn` is a raw stream (could be http/https/tcp/ws)
2. decode the raw stream if PROXY is there.
3. then pass to mqtt parser to read mqtt packets

And he would like it to be:

1. have a connection factory and decode `conn` to be a plain mqtt stream
2. pass `conn` to aedes handle
3. then pass to mqtt parser to read mqtt packets.

I'm ok for boths but I would like to know what you prefer so we can decide if implement a 'websocket module' or a 'connection factory' module

mcollina:
Decoding of PROXY must happen _before_ decoding the websocket protocol. From my understanding the approach @gnought is suggesting will not work with the most common proxies (haproxy for example).

mmm8955405:
I hope that the code will be submitted as soon as possible and provide examples for use. If you can't use an API similar to websocket stream on the server, what is the function of the stream API you provide? However, I can only choose websocket stream at present

robertsLando:
> However, I can only choose websocket stream at present

@mmm8955405 If you are not using protocol-decoder (aedes behind a proxy), you can replace `websocket-stream` in this way: 

```js
const { createWebSocketStream, Server } = require('ws')

function createWebsocketServer (server, broker) {
  const ws = new Server({ server })

  ws.on('connection', function (conn, req) {
    var stream = createWebSocketStream(conn)
    stream._socket = conn._socket
    broker.handle(stream, req)
  })
}

// ... your code here ...

// Than use this function to bind aedes to websocket server
createWebsocketServer(server, broker)
```

gnought:
@mcollina I am suggesting like the following: 
1. have a connection factory and decode conn to be a plain mqtt stream
2. pass conn to aedes handle
3. then pass to mqtt parser to read mqtt packets.

The connection factory is like a layer on top of aedes, if we want ws,
`ws => aedes.handle`
if want want PROXY+ws
`PROXY => ws => aedes.handle`

I will try to create a `createServer()` function to demonstrate what it will look like.

robertsLando:
@gnought @mcollina ping

mcollina:
> Decoding of PROXY must happen _before_ decoding the websocket protocol.

I do not have anything to add to the above. PROXY should not be implemented as a chain of streams.
PROXY is decoded and then the data does not touch that code anymore. None of this should be relevant to Websockets.

robertsLando:
I need to know what @gnought wants to do so

gnought:
sorry, too much work on my shoulder.
I have tried ws over http/2, it works without PROXY only, and requires several hacks in https://github.com/websockets/ws project. I will create a PR for support PROXY on plain mqtt, mqtt over tls and ws over http/https. Tested successfully on haproxy+aedes, maybe I haven't enough time working on unit tests. Help is welcome. :)

demian85:
Hello, I found an issue after upgrading from 0.40 to 0.42. Typescript fails with the following example found in README
```ts
const aedes = require('./aedes')()
const httpServer = require('http').createServer()
const ws = require('websocket-stream')
const port = 8888

ws.createServer({ server: httpServer }, aedes.handle) // Argument of type '(stream: Connection) => Client' is not assignable to parameter of type '() => void'.ts(2345)

httpServer.listen(port, function () {
  console.log('websocket server listening on port ', port)
})
```

Any ideas on how to fix this? does that handle return a Client or void as expected by websocket-stream?

robertsLando:
@demian85 Fixed here: https://github.com/moscajs/aedes/issues/487

Feel free to fix the example in Readme

robertsLando:
@mcollina This ended up with nothing done. I would like to sort this out but I need to know what in your opinion is the best approach.

Actually we use this approach:

```js
const { createWebSocketStream, Server } = require('ws')

function createWebsocketServer (server, broker) {
  const ws = new Server({ server })

  ws.on('connection', function (conn, req) {
   // PROTOCOL DECODE SHOULD BE DONE HERE ?
    var stream = createWebSocketStream(conn)
    stream._socket = conn._socket
    broker.handle(stream, req)
  })
}

const http = require('http')
const server = http.createServer()
// Than use this function to bind aedes to your server
createWebsocketServer(server, broker)
```

So actually we firstly create the websocket strem then we pass it to aedes handle andn [here] we decode the protocol (proxy or what else) and [here](https://github.com/moscajs/aedes/blob/master/lib/client.js#L79-L84) we pass the decoded data to mqtt parser.

So for what I have understanded the best option could be a `serverFactory` like https://github.com/moscajs/aedes/pull/413#issuecomment-584479614 and move the protocol decoder there? Am I correct?

mcollina:
I think so, yes.

robertsLando:
@getlarge Would you like to contribute to that? I can init the repos than there should be some tests to do with your protocol decoder

getlarge:
@robertsLando Sure, i could give it a try next week.
I have to read that feed of info and refresh my memory first, but in the meantime you can start the repo.

robertsLando:
@getlarge Nice! Will start working to this next week: https://github.com/moscajs/aedes-server-factory

getlarge:
@robertsLando Just to be clear what's the plan ?
From the above conversation, i understood that this module was about creating a wrapper around [ws](https://github.com/websockets/ws), but i see in the description of the new repo `Supports tcp, tls, http, https, http2, ws, wss and proxy decoders`. If it's the latter, you think of something like @gnought proposed with the serverFactory ?

As for moving the  protocol decoder step in that module, even though i agree, it seems quite tricky as we need to use the same conn stream that aedes is using to extract the information. Eventually we could use a transform stream in between (before aedes.handle is called) to get the connection infos and remove the proxyprotocol header from the stream, but then where the information about the socket and / or proxy would be stored ?  Should they be returned by the serverFactory and that becomes user responsability ?

And btw, just to add info about the PROXY protocol, it really make sense when we use simple TCP connections (with net module for example), and we need to retrieve information about the `original` address when using a proxy that passes the stream, that's the main use case i had in mind when creating the protocol decoder.
With HTTP we can easily use headers to attach infos from that PROXY header at the server level ( HAProxy, Nginx ... ), so i considered that req.headers were enough to get connection details.

robertsLando:
> If it's the latter, you think of something like @gnought proposed with the serverFactory ?

Yes, that's the main idea. 

> it seems quite tricky as we need to use the same conn stream that aedes is using to extract the information

> but then where the information about the socket and / or proxy would be stored ? Should they be returned by the serverFactory and that becomes user responsability

I started implementing the server factory and I was having the same questions. I'm stucked and I need some ideas... So the main reason of serverFactory would be to simplify the process of creating and attacing aedes to a server and also decode proxy or whatelse, the problem here is that with the actual protocol decoder implementation you had the client instance to attach the connection details and now you don't. I'm stucked :confused: 

@mcollina Ideas?

getlarge:
@robertsLando The client presence in the arguments of protocolDecoder is one thing that we can workaround quite easily by letting the user, append the extracted data her/himself.
What seems more tricky is to start reading the conn stream to check if there is a PROXY protocol header, without messing with aedes that waits for the `full` stream.

I tried something for the serverFactory, it's probably not working but it gives an idea

```js
const WebSocket = require('ws');
const tls = require('tls');
const http = require('http');
const https = require('https');
const http2 = require('http2');
const net = require('net');
const net = require('net');
const protocolDecoder = require('aedes-protocol-decoder');
const { Transform, PassThrough } = require('stream');

function createServer(options, aedesHandler) {
  return new Promise((resolve, reject) => {
    if (!options) {
      return reject(new Error('Missing options'));
    } else if (!aedesHandler) {
      return reject(new Error('aedes handler'));
    }

    var server = null;
    if (options.serverFactory) {
      server = options.serverFactory(aedesHandler, options);
      resolve({ server });
    } else if (options.tls) {
      server = tls.createServer(options, aaedesHandler);
      resolve({ server });
    } else if (options.ws) {
      if (options.https) {
        if (options.http2) {
          server = http2.createSecureServer(options.https);
        } else {
          server = https.createServer(options.https);
        }
      } else {
        if (options.http2) {
          server = http2.createServer();
          server.on('session', sessionTimeout(options.http2SessionTimeout));
        } else {
          server = http.createServer();
        }
      }
      const ws = new WebSocket.Server({ server });
      ws.on('connection', function (conn, req) {
        const stream = WebSocket.createWebSocketStream(conn);
        stream._socket = conn._socket;
        // extract proxy if any ?
        // const removeProxy = removeFromStream(proxyLength);
        // stream.pipe(removeProxy)
        aedesHandler(stream, req);
        resolve({ server });
      });
    } else {
      server = net.createServer(options, (conn) => {
        if (options.useProxy) {
          const rewindableStream = conn.pipe(new Rewindable());
          const { buf, protocol } = extractProxy(rewindableStream);
          const removeProxy = removeFromStream(buf.length);
          const rewound = rewindableStream.rewind();
          conn = rewound.pipe(removeProxy);
        }
        aedesHandler(conn);
        resolve({ server });
      });
    }
  });
}

const extractProxy = (conn, req) => {
  const buf = conn.read(null);
  const protocol = protocolDecoder({ req }, buf);
  // proto contains {data, ipAddress, port, serverIpAddress, isProxy}
  return { buf, protocol };
};

class Rewindable extends Transform {
  constructor() {
    super();
    this.accumulator = [];
  }

  _transform(buf, enc, cb) {
    this.accumulator.push(buf);
    callback();
  }

  rewind() {
    const stream = new PassThrough();
    this.accumulator.forEach((chunk) => stream.write(chunk));
    return stream;
  }
}

const removeFromStream = (length) => {
  let buffer = Buffer.alloc(0);
  let removed = false;

  return new Transform({
    transform(chunk, encoding, callback) {
      if (!removed || buffer.length <= length) {
        buffer = Buffer.concat([buffer, chunk], buffer.length + chunk.length);
        const partChunk = buffer.slice(length);
        this.push(partChunk);
        removed = true;
      } else {
        this.push(chunk);
      }
      callback();
    },
  });
};

module.exports = {
  createServer,
};
```

robertsLando:
It seems ok but also a little bit tricky and maybe performance killer? Also If you check [here](https://github.com/moscajs/aedes/blob/master/lib/client.js#L76) the protocol decoder was running only in case connack wasn't sent, is there any reason to do it on every rquest?

getlarge:
It's based on Promise, so that we can return the server + the connDetails to the user.
Check the else condition that uses the net.createServer, i only worked on this case now.
Based on the idea that we should extract connection details and eventual Proxy header:
- we start to read the conn stream (wrapped in a transform stream),
- use the buffer to extract informations needed
- `rewind` the stream
- remove the proxy part from the rewinded stream (if any) with another transform stream
- use that stream chains in aedes.handle

The advantage is that aedes would not have to bother with protocol decoding and that would allow to remove those ugly nested conditions in `nextBatch` ( that i made :) )

getlarge:
Previously we needed that connack check because nextBatch was executed for each incoming packet, in the serverFactory, it would be executed only on connection event

robertsLando:
@getlarge I think that this could be the way, good job! Would you like to submit a PR to server factory? I also would like to know if @mcollina thoughts about this solution and If this caould be improved someway :)

PS: If you are interested I could add you to our org MoscaJS org

mcollina:
as long as it does not slow performance, I'm good.

robertsLando:
@mcollina Do you think that solution could slow it or not? @getlarge for this we will add some performance test to see how it goes

getlarge:
@robertsLando Yes i can make a PR soon, but i'll have to change the protocolDecoder implementation first, and yes for the org, i will not have not to handle some forks anymore ;)

@mcollina For the perf, the connection handling will most probably slower if there is a proxy headers because of the stream manipulation, but it could only affect positively the common operations ( incoming messages handling ), as we could remove the nested conditions in aedes.client.nextBatch.
But we can add some tests to be sure.

PR #414, opened 2020-02-07
Added test messageId should reset to 1 when reach 65535

gnought:


lib/qos-packet.js
```diff
@@ -10,7 +10,7 @@ function QoSPacket (original, client) {
 
   if (!original.messageId) {
     this.messageId = client._nextId
-    if (client._nextId === 65535) {
+    if (client._nextId >= 65535) {
```
mcollina: why? could this happen?

gnought: depends, it prevents someone set `_nextId` by themselves.

PR #416, opened 2020-02-07
Removed obsolete tests

gnought:
Those tests have been moved into aedes-protocol-decoder repo


robertsLando:
In case of' websocket-stream updated to ws' those were useful to check there was breaking changes on protocol-decoder. Like the `_socket` property that wasn't present

gnought:
Please test it in aedes-protocol-decoder.

robertsLando:
Would also decrease coverage, i think we need at least one test that uses decode protocol

Daniel

robertsLando:
@gnought I suggest to create a dummy protocol decoder test that simply parse data

https://github.com/moscajs/aedes/blob/master/lib/client.js#L80

PR #421, opened 2020-02-10
Coverage

robertsLando:


test/will.js
```diff
@@ -113,6 +113,24 @@ test('delivers old will in case of a crash', function (t) {
...
+  })
+  t.tearDown(broker.close.bind(broker))
+
+  var brokerId = 'dummyBroker'
+
+  broker.brokers[brokerId] = Date.now() - heartbeatInterval * 3.5
+
+  setTimeout(() => {
+    t.equal(broker.brokers[brokerId], undefined, 'Broker deleted')
+  }, heartbeatInterval * 4)
```
mcollina: can we use http://npm.im/lolex to test this?

robertsLando: Why? BTW lolex is deprecated in favor of https://github.com/sinonjs/fake-timers

mcollina: timers are highly unreliable in CI.

robertsLando: I don't really understand how they work. I have tried to simply add 

```js
require("@sinonjs/fake-timers").install()
```

To `test/helper.js` but seems timers are not working. I have also tried with other methods but I can't make them work

robertsLando:
@mcollina Are you ok if we land this and we do another one for timers fix?

mcollina:
I would prefer to fix the timers.

robertsLando:
@mcollina Is it ok?

Issue #428, opened 2020-02-12
[bug] Catch mqtt-packet parser errors


robertsLando:
@mcollina Actually we don't catch mqtt-pacekt parser errors. I have tried to create a test sending wrong cmd to the inStream and mqtt-packet throws an error `Unknown command` that is not catched even if I add [here](https://github.com/moscajs/aedes/blob/master/lib/client.js#L55)

```js
  this._parser.on('error', onError)
```

on client.js.

mcollina:
That’s ok and it should destroy the connection
Il 12 feb 2020, 11:37 +0100, Daniel Lando <notifications@github.com>, ha scritto:

robertsLando:
@mcollina It's not ok, it doesn't catch the error

mcollina:
So the error is not emitted? Can you upload a test to reproduce?

robertsLando:
@mcollina 

```js
test('catch error mqtt-packet parse error', function (t) {
  t.plan(1)

  const s = noError(connect(setup()), t)
  t.tearDown(s.broker.close.bind(s.broker))

  s.broker.on('clientError', function () {
    t.pass('error thrown')
  })

  s.broker.on('clientReady', function (c) {
    s.inStream.write({ cmd: 'pippo' })
  })
})
```

robertsLando:
Here: https://github.com/moscajs/aedes/runs/440829422#step:6:384

mcollina:
That error is thrown by the test harness when it generates a packet and not by Aedes.

PR #429, opened 2020-02-12
Extend coverage to 100%

robertsLando:
#428

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -53,7 +53,6 @@ function Client (broker, conn, req) {
   this._parser.client = this
   this._parser._queue = [] // queue packets received before client fires 'connect' event. Prevents memory leaks on 'connect' event
   this._parser.on('packet', enqueue)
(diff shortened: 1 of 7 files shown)


gnought:
@robertsLando We can't catch the error, as mqtt-parser has already filtered out all unknown verb, but It is safe to have a check in `lib/handlers/index.js`. That what I include in the next branch https://github.com/gnought/aedes/blob/next/test/basic.js#L293

robertsLando:
With this we should be 100% covered by tests :tada:

robertsLando:
@gnought Could you review this?

gnought:
@robertsLando I have checkout the coverage branch, but I can't get 100% coverage by `npm run unit:report`

robertsLando:
@gnought It give 100% to me. Check https://coveralls.io/builds/28705089

robertsLando:
@mcollina Could you review this?

Issue #373, opened 2020-01-26
Topic matching: Single-level wildcard does not match empty level


ffa500:
I'm running aedes@0.40.1. My MQTT.JS v3.1.1 client subscribes to topic "a/+/c" but it doesn't receive publications on topic "a//c". Publishing on "a/b/c" is okay. Same issue with "a/+": matching against "a/" does not work, but "a/b" is fine.

Seems aedes is not supporting single-wildcard matching against empty levels. Mosca does not support it either. Yet, other brokers like mosquitto or HiveMQ support this case.

MQTT Spec v3.1.1 is not excluding this special case in its definition of the + wildcard:

"The single-level wildcard can be used at any level in the Topic Filter, including first and last levels. Where it is used it MUST occupy an entire level of the filter [MQTT-4.7.1-3]. It can be used at more than one level in the Topic Filter and can be used in conjunction with the multilevel wildcard."

It also states that  "Adjacent Topic level separators indicate a zero length topic level", so I assume empty levels are valid and should also match the + wildcard.

Thanks for taking a look at this issue!

mcollina:
would you mind to send a PR to address this? Or at least point us in the right direction.
Il 26 gen 2020, 13:36 +0100, Hubertus Hohl <notifications@github.com>, ha scritto:

gnought:
This relates to Qlobber used in our mqemitter.
Qlobber (https://github.com/davedoesdev/qlobber) states
> {String} wildcard_one The character to use for matching exactly one non-empty word in a topic. Defaults to '*'. MQTT uses '+', for example.

`wildcard_one` only matches non-empty words, there are 2 options.
1. submit a PR to Qlobber
2. or tweak mqemitter to add one match (e.g. a//c and a/)

@mcollina which one is preferred, or any good options?

mcollina:
cc @davedoesdev what do you think?

davedoesdev:
I'm happy to add an option to Qlobber to change the behaviour so `wildcard_one` matches empty word as well. Is that the cleanest solution for aedes?

robertsLando:
@davedoesdev I think that would be good, or maybe simply add an option like `matchEmptyLevels` , to be sure that this doesn't break any existing instance. @mcollina What do you think it's best?

mcollina:
+1 for a new option!
Il 29 gen 2020, 08:08 +0100, Daniel Lando <notifications@github.com>, ha scritto:

davedoesdev:
[...]

PR #433, opened 2020-02-19
Fix: Single-level wildcard match empty level

gnought:
Fix https://github.com/moscajs/aedes/issues/373


robertsLando:
I suggest to add some info on readme for the ones using custom mqemitters about that option for empty level

Daniel

gnought:
Or we could do it better. Check custom mqemitters if they inherit base mqemitter, and set `this.mq.matchEmptyLevels = true` directly.

robertsLando:
I would prefer changing the defaults, most users would miss that for sure

gnought:
Do all of you mean change the `matchEmptyLevels` default to true in `mqemitter` package?

mcollina:
I've released mqemitter v4.1.1 with the flag turned to true.

Il giorno mer 19 feb 2020 alle ore 10:19 Gnought <notifications@github.com>
ha scritto:

aedes.js
```diff
@@ -44,7 +44,10 @@ function Aedes (opts) {
   this.counter = 0
   this.queueLimit = opts.queueLimit
   this.connectTimeout = opts.connectTimeout
-  this.mq = opts.mq || mqemitter(opts)
+  this.mq = opts.mq || mqemitter({
+    concurrency: opts.concurrency,
+    matchEmptyLevels: true // [MQTT-4.7.1-3]
```
robertsLando: Should this be removed? It should be automatically turned on now

gnought: It is safe to set it as we know it.

Issue #228, opened 2019-01-25
Support of  Shared Subscriptions


MrSpark2591:
I was reading article and found out [this article](https://www.hivemq.com/blog/mqtt-client-load-balancing-with-shared-subscriptions/)
Can it be done in this package ?
Or is there any other way we can achieve such scenario?

mcollina:
Not at this point, you’ll have to code/provide that yourself.

MrSpark2591:
I found out that we will need multiple changes for the support.

1) Multiple client session. currently only one client is allowed to connect to broker with that client id.
2) Mqtt.js support of v5 currently [under development](https://github.com/mqttjs/MQTT.js/issues/651).

I wil try if i can get any work around for this. let me know if i am missing anything here.

mcollina:
The work on aedes is not in progress at all.

MrSpark2591:
Is it due to support of mqtt or you need more maintainer to get it going? if anyways I can help let me know.

mcollina:
I have no time to implement this at the moment.

abhargav7:
any work around you guys found on it? That would be helpful Thanks!!

robertsLando:
@gnought @mcollina This seems not that hard to implement, it's also part of mqtt5 but could be implemented without having mqtt5 support.

abhargav7:
Thanks great to hear its quick. Its a powerful feature to have and solves lots of complex problems.

robertsLando:
@abhargav7 I don't think it will be quick, it may require also an update of aedes-persistence

abhargav7:
Sure, Thanks, when do you think you can contribute this for the betterment of aedes broker?

robertsLando:
I have firstly to speak about this with others maintainers

abhargav7:
Thanks, please!

gnought:
I think we need a custom mqemitter.

robertsLando:
Nope, we just need to create a sharedGroup client that keeps track of all clients subscribed to the group and sends the updates to only one client following round robin algorithm 

Daniel

PR #437, opened 2020-02-20
WIP: shared subscriptions

robertsLando:
This is a work in progress for the feature. I would like to know your opinion about this idea and if this could be the right direction or not. @gnought @mcollina 

#228

### DIFF
--- a/aedes.js
+++ b/aedes.js
@@ -26,6 +26,7 @@ const defaultOptions = {
   authorizeSubscribe: defaultAuthorizeSubscribe,
   authorizeForward: defaultAuthorizeForward,
   published: defaultPublished,
+  sharedTopics: false,
   trustProxy: false,
   trustedProxies: [],
   queueLimit: 42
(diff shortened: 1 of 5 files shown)


robertsLando:
@mcollina So what should be the way? Suggestions?

mcollina:
It needs to be architected/investigated, I’m basically stating requirements. I currently do not have time for doing so in the near future.

A practical way is to have a “plugin” built on top of Redis Streams. This might have significant limitation related to how #/+ work.

A more complex solution is to publish the “active” shared subscription for each peer in the broker metadata, and have the message being routed to a different broker if the current one does not have a an active shared subscription for that topic.

robertsLando:
Could my actual implementation work in a single-instance model at least? 

About your answer I think the best way would be to add shared subscriptions to aedes-persistence or it becomes hard to handle. I was thinking about a solution using `$SYS` topics but I think it becomes complex and has very bad efficiency

gnought:
good try.
imo. aedes is a multi process model, we need to make it flexile rather than implemented in single-instance.

robertsLando:
@gnought Could we speak about a possible solution for this? I think a possible soulutions could be:

Once we receive a shared subscriuption we add the subscription to persistence with a `created` date (set to now) an a `lastUpdate` date set to null. Before each delivery the client will check in the persistence for the shared subscription to that topic with the lower lastUpdate and if the subscription returned has his clientId he delivers the packet and update his subscription `lastUpdate`

Should this work? Thoughts?

gnought:
There may be more work to do before that, and make the whole picture be clear. Would you mind to extract out interfaces that other aedes persistence should implement to, it is nicer if we have a typing in aedes-persistence.

robertsLando:
@gnought Check https://github.com/moscajs/aedes-persistence/pull/48 Than I will add new methods

robertsLando:
@gnought I close this and start working on the new one

abhargav7:
@robertsLando  does closing means you are creating a PR from new branch sooner, or case is closed :)
Thanks for supporting this.

PR #444, opened 2020-02-22
Enhanced retain.js tests by fake timers

gnought:


test/retain.js
```diff
@@ -167,7 +168,9 @@ test('reconnected subscriber will not receive retained messages when QoS 0 and c
 test('new QoS 0 subscribers receive QoS 0 retained messages when clean', function (t) {
   t.plan(9)
 
+  const clock = Faketimers.createClock()
```
mcollina: You’d probably need to uninstall the fake timer

gnought: As not use `install()`, we cannot call `uninstall()` to reset the clock. 
I put `clock.reset()` for it.

gnought:
@mcollina looks good?

PR #445, opened 2020-02-22
Add a non-blocking concurrent connection test

gnought:


test/not-blocking.js
```diff
@@ -6,6 +6,50 @@ const net = require('net')
 const Faketimers = require('@sinonjs/fake-timers')
 const aedes = require('../')
 
+test('connect 500 concurrent clients', function (t) {
+  t.plan(3)
+
+  const broker = aedes()
+  const server = net.createServer(broker.handle)
+  const total = 500
```
mcollina: can we just do 100 instead? 500 might fail on some OS

gnought: Someone posted an issue said aedes stop receiving new connections when reaching 200. Could we raise to 250 for having a proof?

mcollina: l think their problem was due to the ulimits on their system rather than Aedes. I would like Aedes tests to pass on a vanilla system (200 is ok, ad 256 is the limit on Mac).

There is nothing we can prove on our side for these errors.

gnought: Let make it to 200. looks good to you?

PR #446, opened 2020-02-22
Refactored EventEmitter

gnought:


gnought:
imo, It is simpler than using `const EE = require('events').EventEmitter`
Maybe I miss the differences. Would you mind to point it out if so? :)

gnought:
And I follow the example in node doc https://nodejs.org/api/events.html#events_class_eventemitter

PR #454, opened 2020-02-26
Handle unsubscription safer while client closes

gnought:
We're better not to inject a custom key `close` into one of argument in `handleUnSubscribe` fn. Malicious unsubscribe packet may break aedes unsubscribe logic.


gnought:
@mcollina we already cover it in https://github.com/moscajs/aedes/blob/209efd018ab8645f061aed50c0dacf279bb51e6b/test/meta.js#L184-L204

robertsLando:
@gnought I think he means a test where a malicious packet with close prop is received

gnought:
ping @mcollina

PR #455, opened 2020-02-26
Added test on unrecognised params in sub packet structure

gnought:
In https://github.com/moscajs/aedes/pull/408, we remove the `restore` extra key in subscription packet structure, not to break aedes logic.

### DIFF
--- a/test/meta.js
+++ b/test/meta.js
@@ -154,7 +154,7 @@ test('emit subscribe event', function (t) {
 })
 
 test('emit subscribe event if unrecognized params in subscribe packet structure', function (t) {
-  t.plan(2)
+  t.plan(3)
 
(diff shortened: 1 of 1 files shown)


gnought:
ping @mcollina

test/meta.js
```diff
@@ -153,6 +153,26 @@ test('emit subscribe event', function (t) {
...
+
+  broker.on('subscribe', function (subscriptions, client) {
+    t.equal(subscriptions, subs)
+    t.deepEqual(client, s.client)
+  })
+
+  s.client.subscribe({
+    subscriptions: subs,
+    restore: true
+  }, function () {})
```
mcollina: Please add an assertion within this function

PR #456, opened 2020-03-03
Expose protocol version in Client object

gnought:
as title, we expose `CONNECT` protocol version into Client that add flexibility to our users to fine-tune their hook handlers.

### DIFF
--- a/lib/handlers/connect.js
+++ b/lib/handlers/connect.js
@@ -81,6 +81,7 @@ function init (client, packet, done) {
 
   client.id = clientId || 'aedes_' + shortid()
   client.clean = packet.clean
+  client.version = packet.protocolVersion
   client._will = packet.will
 
   client.broker._series(
--- a/aedes.d.ts
+++ b/aedes.d.ts
@@ -74,6 +74,7 @@ declare namespace aedes {
   interface Client extends EventEmitter {
     id: string
     clean: boolean
+    version: number
     conn: Connection
     req?: IncomingMessage
     connecting: boolean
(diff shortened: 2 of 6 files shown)


lib/handlers/connect.js
```diff
@@ -81,6 +81,7 @@ function init (client, packet, done) {
 
   client.id = clientId || 'aedes_' + shortid()
   client.clean = packet.clean
+  client.version = packet.protocolVersion
```
GavinDmello: @gnought Looks good overall. I would be inclined towards changing the property from `version` to `protocolVersion` as it could easily be confused with the library version of the client rather than the protocol version, for those who don't read the documentation.

robertsLando: @GavinDmello I think that if someone would like to get the library version it should get it from aedes and not a Client

GavinDmello: Oh. I meant, client.version sounds like the library version of the client which is connecting, which could be different for each each client. Don't think that should be a Aedes/broker level property or a property in our code at all.

gnought: The concern from @GavinDmello comes to my mind when I go through the design process. It sounds like it's a library version but the point @robertsLando is correct, `Client` is a part of aedes module, not a library. `client.version` is designed to tell "this is MQTT 3.1 client, or MQTT 3.1.1 client or MQTT 5 client"
If we choose the name `client.protocolVersion`, it it weird that telling "hey what protocolVersion are you" where the client is not a packet.

GavinDmello: That was a suggestion. Happy for this to go ahead.

robertsLando:
@gnought Could you merge this?

gnought:
Sure

PR #457, opened 2020-03-03
Updated dependencies

gnought:


package.json
```diff
@@ -83,26 +83,26 @@
     "snazzy": "^8.0.0",
     "standard": "^14.3.1",
     "tap": "^14.10.6",
-    "typescript": "^3.8.2",
+    "typescript": "^3.8.3",
     "websocket-stream": "^5.5.0"
   },
   "dependencies": {
     "aedes-packet": "^2.3.1",
```
robertsLando: Could you remove `aedes-packet` as dep? You can now get it from `aedes-persistence`

gnought: That will make aedes core depends on the Packet export in aedes-persistence / inherited packages. Will it be fault-tolerance if someone made their own standalone packages and plug into aedes?

robertsLando: In my latest commits on aedes-cached-persistence and aedes-persistence I spoke about this with @mcollina, all pakcages that use aedes-persistence will also use aedes-packet so he also suggested to completely remove aedes-packet module and place it in aedes-persistence. This is a good option IMO because we don't have to keep it updated across all moscajs modules that are using it

robertsLando: @mcollina ping

mcollina: I don't care too much about this. IMHO it makes aedes-packet superfluous as a package, but @robertsLando is very convinced this is needed so here we go.

gnought: @robertsLando I will create another PR for removing aedes-packet dep.

PR #468, opened 2020-04-09
Return granted qos in broker subscribe event and better subAck docs

### LINKED ISSUE #467: [bug]subscribe Event will triggered while topic is negated.
**System Information**
- Aedes: 0.41.0
- NodeJS: 10.15.3
- OS: Centos 7

**Describe the bug**
Subscribe Event will triggered while topic is negated.

**To Reproduce**
```
//server.js
const aedes = require('aedes')();

aedes.authorizeSubscribe = function (client, sub, callback) {
    if (sub.topic === 'presence') {
        sub = null
    }
    callback(null, sub)
}

aedes.on("subscribe", (subs, client)=>{
    console.log(`subscribe event`, subs) // triggered
})


const server = require('net').createServer(aedes.handle)
const port = 1883

server.listen(port, function () {
    console.log('server started and listening on port ', port)
})
```

```
//client.js
const mqtt = require('mqtt')
let client  = mqtt.connect({
    servers: [{
        host: 'localhost',
        port: 1883,
        protocol: 'tcp'
    }]
});

client.on('connect', function () {
    client.subscribe('presence', function (err, granted) {
        if (!err) {
            console.log(granted) //[ { topic: 'presence', qos: 128 } ]
        }else
            console.log(err)
    })
});
```

robertsLando:
Fixes #467

### DIFF
--- a/docs/Aedes.md
+++ b/docs/Aedes.md
@@ -369,6 +369,8 @@ aedes.authorizeSubscribe = function (client, sub, callback) {
 }
 ```
 
+__ATTENTION__: When a subscription is negated, your client will receive back a SubAck with `qos: 128` that means `Failure` and your client will never receive packets published to that topic.
+
 ## Handler: authorizeForward (client, packet)
 
 - client: [`<Client>`](./Client.md)
--- a/docs/Aedes.md
+++ b/docs/Aedes.md
@@ -138,7 +138,7 @@ Emitted when `client` sends a `PINGREQ`.
 
 Emitted when `client` successfully subscribe the `subscriptions` in server.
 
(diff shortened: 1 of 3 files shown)


robertsLando:
> This needs unit tests

I was waiting for this :smile: 

Just to know if this approach is fine for you, if so I will go on with tests

mcollina:
I don't know, it's hard to reason about without tests.

gnought:
If someone intends to duplicate topic in its subscription, why we need to dedupe them in the subscription event?

gnought:
And we need to define what the purpose of “subscribe” events do. If we wanna acknowledge back subscription successful status by modifing qos, say 128, we need to refine the “subscribe” event doc too.

robertsLando:
> If someone intends to duplicate topic in its subscription, why we need to dedupe them in the subscription event

Why someone would do this?

> And we need to define what the purpose of “subscribe” events do

I suggest to return both original subs and result subs

PR #474, opened 2020-04-22
feat: emptyOutgoingQueue

### LINKED ISSUE #473: [question] How do I properly wipe a disconnected session?
Hi!

Given that MQTTv5 session expiry is not implemented yet, what is the current best way to wipe a persistent (non-clean) session that's been disconnected for a while? 

E g, if a persistent session has been disconnected for a specific time interval, or the number of queued messages has exceeded a certain number, or something else, I want aedes to completely forget about the client/session, both its subscriptions and its current queue of offline messages.

I'm not sure if this is done in aedes or directly in the persistence layer, but if the latter, let's assume that I'm using `aedes-persistence`, i e, memory only persisence.

Thanks in advance!

robertsLando:
This method allows to empty the client outgoing queue based on 

Fixes #473


diwic:
@robertsLando Looks awesome!

Combine it with a `Client.prototype.expireSession` function that calls `emptyOutgoingQueue` after `close` and you've implemented part of MQTTv5 session expiry, right?

robertsLando:
> Combine it with a Client.prototype.expireSession function that calls emptyOutgoingQueue after close and you've implemented part of MQTTv5 session expiry, right?

Yes but needs some more work, just a little step forward :)

robertsLando:
I also think that this could be done in a better way by adding a new method to persistence `emptyOutgoingQueue` that takes as argument only the client id and deletes all outgoing messages related to that clientId. 

Thoughts @mcollina @gnought ?

lib/client.js
```diff
@@ -11,6 +11,8 @@ const QoSPacket = require('./qos-packet')
 const handleSubscribe = require('./handlers/subscribe')
 const handleUnsubscribe = require('./handlers/unsubscribe')
 const handle = require('./handlers')
+const through = require('through2')
```
mcollina, line 14: You can just use a `new Transform({ transform() {} }})` instead.

robertsLando: What are the benefits?

robertsLando: Ok fixed, was missing `objectMode: true`. Transforms are not really well documented on Nodejs docs, cannot find a reference on transform options :confused:

robertsLando: Should I remove through2 as dep and use Transform instead in all project? Has it better performance?

robertsLando:
Should I remove `through2` as dep and use `Transform` instead in all project? Has it better performance?

PR #479, opened 2020-04-29
Bump deps and fixed tests

robertsLando:
@mcollina 

I don't know if this was wanted or no. In mqtt@3.0.0 the only stream error thrown was 'ECONNREFUSED' but in  mqtt@4.0.0 now it throws all this errors:

```js
var socketErrors = [
  'ECONNREFUSED',
  'EADDRINUSE',
  'ECONNRESET',
  'ENOTFOUND'
]
```
Source: https://github.com/mqttjs/MQTT.js/pull/1076/files#diff-50cfa59973c04321b5da0c6da0fdf4feR30

Here you see the commit that has changed this: https://github.com/mqttjs/MQTT.js/pull/1076/files#diff-50cfa59973c04321b5da0c6da0fdf4feL325


README.md
```diff
@@ -133,6 +167,51 @@ Average Bandwidth (msg/sec): 28.115
 Total Bandwidth (msg/sec):   28114.678
 ```
 
+#### Redis Persistence and Redis Emitter - With Clusters
```
mcollina, line 170: How many servers in the cluster?

README.md
```diff
@@ -133,6 +167,51 @@ Average Bandwidth (msg/sec): 28.115
 Total Bandwidth (msg/sec):   28114.678
```
mcollina, line 167: was this run on the same server?

robertsLando: Same yes

README.md
robertsLando: one for each core, so 8

robertsLando:
About your LGTM @mcollina, I don't understand if the answer to my question

> I don't know if this was wanted or no. In mqtt@3.0.0 the only stream error thrown was 'ECONNREFUSED' but in mqtt@4.0.0 now it throws all this errors:

Is yes it is wanted or not it should not work like that

mcollina:
tests are passing and code is ok for me. I didn’t understand your question.

robertsLando:
@mcollina I don't think mqtt client should throw that error when closing client. That's why my question

I think that once the stream is closed using client end there should be no error thrown by the stream

mcollina:
I don’t know and it does not matter for Aedes. There are a few new folks that are working on MQTT.js.

PR #486, opened 2020-05-13
proposed fix #469

### LINKED ISSUE #469: subscribe/unsubscribe processed out of order when persistence is used
When a client subscribes and unsubscribes repeatedly to the same topic in quick succession, sometimes these are processed out of order, resulting in the client believing it has subscribed but it is actually not subscribed.

It looks like the root cause is that the subscribe handler may [complete synchronously](https://github.com/moscajs/aedes/blob/295538898deaa57edcd6f62c469a55dd02a3a58c/lib/handlers/subscribe.js#L104) but the unsubscribe handler always [completes asynchronously](https://github.com/moscajs/aedes/blob/295538898deaa57edcd6f62c469a55dd02a3a58c/lib/handlers/unsubscribe.js#L30), so it's possible for the unsubscription to complete after the subscription.

As a workaround, I have implemented a modified persistence which synchronously completes the unsubscribe call when the client is "clean".

aquark:


lib/handlers/subscribe.js
```diff
@@ -100,7 +100,8 @@ function storeSubscriptions (sub, done) {
 
   const packet = this.packet
   const client = this.client
-
+  // for correctness it's essential that if subscriptions are synchronous,
```
robertsLando, line 103: You could add a reference to the issue here

aquark:
> Can you please add a unit test?

Sorry, I can't dedicate any more time to this issue.

robertsLando:
I can write a test for this if needed

---

Daniel - Software Engineer

Support me at:
    Github sponsors

robertsLando:
Close this in favor of #488

PR #488, opened 2020-05-14
fix: remove subs only when clean flag is false

### LINKED ISSUE #469: subscribe/unsubscribe processed out of order when persistence is used

robertsLando:
Fixes #469

### DIFF
--- a/test/client-pub-sub.js
+++ b/test/client-pub-sub.js
@@ -514,6 +514,36 @@ test('unsubscribe a client', function (t) {
   connect(setup(broker))
 })
 
+test('unsubscribe should not call removeSubscriptions when [clean=true]', function (t) {
+  t.plan(2)
+
+  const broker = aedes()
+  t.tearDown(broker.close.bind(broker))
+
+  broker.persistence.removeSubscriptions = function (client, subs, cb) {
+    cb(Error('remove subscription is called'))
+  }
+
+  broker.on('client', function (client) {
+    client.subscribe({
+      topic: 'hello',
+      qos: 1
+    }, function (err) {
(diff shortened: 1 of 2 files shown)


robertsLando:
@mcollina ping

test/client-pub-sub.js
```diff
@@ -514,6 +514,36 @@ test('unsubscribe a client', function (t) {
...
 })
 
+test('unsubscribe should not call removeSubscriptions when [clean=true]', function (t) {
+  t.plan(2)
+
+  const broker = aedes()
+  t.tearDown(broker.close.bind(broker))
+
+  broker.persistence.removeSubscriptions = function (client, subs, cb) {
+    cb(Error('remove subscription is called'))
```
mcollina, line 524: this should be a t.error instead

robertsLando: The error is caught by the callback of unsubscribe

PR #492, opened 2020-05-20
fix: clean up the write callbacks in case of error

### LINKED ISSUE #483: [Bug] Aedes stops accepting subscribe/ publish
Hello 

I have migrated from mosca few days back . I have a test server with ~10 clients connected and another production server with ~400 clients connected. The smaller instance works great but the bigger one stops responding to subscribe and publish requests after few days . Not sure what could be the cause. I checked the CPU/ memory usage. Its not high. 

**Aedes Code**
https://pastebin.com/raw/Zbf58pR1

**Logs  - I only see clients being connected but unable to publish / subscribe**
```
authorize :- device-5e3c5054ca15e200c69dd8c6
client connected Tin-fb49
authorize :- device-5e3c5054ca15e200c69dd8c6
client connected Tin-fb49
authorize :- device-5e3csdfsdfsdfsdfsdfsdfsdfsdf
client connected Tin-fb49
authorize :- device-5e3c5054ca15esdfsdfsdfsdfds
client connected Tin-rewr
```

robertsLando:
Fixes #483

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -56,6 +56,7 @@ function Client (broker, conn, req) {
   this._parser.client = this
   this._parser._queue = [] // queue packets received before client fires 'connect' event. Prevents memory leaks on 'connect' event
   this._parser.on('packet', enqueue)
+  this._parser.on('error', noop)
   this.once('connected', dequeue)
 
   function nextBatch (err) {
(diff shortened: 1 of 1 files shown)


robertsLando:
@mcollina I dunno how to test the request drain

robertsLando:
@mcollina This seems to have partially fixed the problem, seems that the problem persists when using authorizePublish/authorizeSubscribe functions could the problem be something similar? some callbacks never called?

robertsLando:
@mcollina Could the problem be caused by async functions? I mean if autorizeFunctions are async? Could this cause some memory leaks?

robertsLando:
I think that it's a user error let's merge this for now

PR #493, opened 2020-05-21
fix: catch writeToStream errors

robertsLando:
@mcollina this should fix unhandled expections when a non-valid packet is write to stream

### DIFF
--- a/lib/write.js
+++ b/lib/write.js
@@ -4,13 +4,17 @@ const mqtt = require('mqtt-packet')
 
 function write (client, packet, done) {
   if (client.conn.writable && (client.connecting || client.connected)) {
-    const result = mqtt.writeToStream(packet, client.conn)
-    if (!result && !client.errored && done) {
-      client.conn.once('drain', done)
-      return
-    }
-    if (done) {
-      setImmediate(done)
+    try {
+      const result = mqtt.writeToStream(packet, client.conn)
+      if (!result && !client.errored && done) {
+        client.conn.once('drain', done)
+        return
+      }
+      if (done) {
+        setImmediate(done)
+      }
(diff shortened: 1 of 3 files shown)


robertsLando:
@mcollina I'm :+1: for the test but why this isn't working? 

```js
// Catch invalid packet writeToStream errors
test('reject client requested for unsupported protocol version', function (t) {
  t.plan(3)

  const broker = aedes()
  t.tearDown(broker.close.bind(broker))

  const s = setup(broker)

  var packet = [
    16, 22, 0, 4, 77, 81, 84, 84,
    4, 2, 0, 0, 0, 10, 109, 121,
    99, 108, 105, 101, 110, 116, 105, 100,
    85, 0
  ]

  s.inStream.write(Buffer.from(packet))

  s.outStream.on('data', function (packet) {
    t.fail('no data sent')
  })
  broker.on('connectionError', function (client, err) {
    t.equal(client.version, null)
    t.equal(err.message, 'Unknown command')
    t.equal(broker.connectedClients, 0)
  })
})
```

I mean, it works as test but it doesn't make that function throw

mcollina:
I mean require the function directly and mock everything.

robertsLando:
@mcollina I tried even that way and I'm not able to reproduce the problem, maybe I'm missing something

mcollina:
You should use proxyquire to mock writeToStream.

robertsLando:
@mcollina I saw that if there are errors or client is no connecting/connected the write never calls `done` is this correct or should we also call `done(err)`?

mcollina:
> I saw that if there are errors or client is no connecting/connected the write never calls done is this correct or should we also call done(err)?

I do not know.

PR #497, opened 2020-05-29
docs: LGTM badges

robertsLando:


robertsLando:
@mcollina LGTM shows this alert: https://lgtm.com/projects/g/moscajs/aedes/alerts/?mode=list

Do you think we can remove that empty buffer as var? I cannot understand why it is needed

mcollina:
sure thing

PR #500, opened 2020-06-03
Add Kuzzle to "Made with aedes" section

Aschen:
Hi!

We recently change mosca for aedes for our MQTT support in Kuzzle and we are very happy of it!


robertsLando:
LGTM for me. @mcollina ?

PR #507, opened 2020-06-12
fix: prevent client connection to be closed when doing connack

robertsLando:
Could be a possible fix of #483

### DIFF
--- a/lib/handlers/connect.js
+++ b/lib/handlers/connect.js
@@ -75,7 +75,6 @@ function init (client, packet, done) {
     doConnack(
       { client: client, returnCode: returnCode, sessionPresent: false },
       done.bind(this, error))
-    client.conn.end()
     return
   }
 
--- a/test/connect.js
+++ b/test/connect.js
@@ -290,9 +290,10 @@ test('reject clients with > 23 clientId length in MQTT 3.1.0', function (t) {
 
   const s = setup(broker)
 
-  var end = s.client.conn.end
+  var conn = s.client.conn
+  var end = conn.end
 
-  s.client.conn.end = function () {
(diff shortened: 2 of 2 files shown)


mcollina:
I would add a test. Also, it seems really obscure on why this is the fix.

robertsLando:
@mcollina Based on [this](https://github.com/moscajs/aedes/issues/483#issuecomment-642880187) comment I was looking at the connect handler and I found `client.conn.end()` immediatly after the `doConnack` call, this could cause some race condition and the connection could be closed before the connack is sent, also it is useless as when the connack is sent done callback is called with an error and it will be catched by client `nextBatch` (https://github.com/moscajs/aedes/blob/master/lib/client.js#L62) and the connection will be closed

robertsLando:
If this seems logic for you now I will write a test for that

lib/handlers/connect.js
```diff
@@ -75,7 +75,6 @@ function init (client, packet, done) {
     doConnack(
       { client: client, returnCode: returnCode, sessionPresent: false },
       done.bind(this, error))
-    client.conn.end()
```
mcollina, line 78: When will the client be disconnected then?

robertsLando: Like I said in previous comment: doConnack will call done with `error`, this will be catched by client (nextBatch)[https://github.com/moscajs/aedes/blob/master/lib/client.js#L62] that will emit error that will call (client.onError)[https://github.com/moscajs/aedes/blob/master/lib/client.js#L92] that will close client connection [here](https://github.com/moscajs/aedes/blob/master/lib/client.js#L192)

PR #508, opened 2020-06-19
fix: write leaks

robertsLando:
Actually [write](https://github.com/moscajs/aedes/blob/master/lib/write.js#L5) doesn't return errors to done callback but insteed it calls `client._onError` that will emit a clientError/connectionError event and disconnect the client. This could cause some memory leaks as when an error occurs the write callback is never called

### DIFF
--- a/test/basic.js
+++ b/test/basic.js
@@ -132,7 +132,7 @@ test('publish empty topic throws error', function (t) {
 })
 
 // Catch invalid packet writeToStream errors
-test('catch write errors', function (t) {
+test('return write errors to callback', function (t) {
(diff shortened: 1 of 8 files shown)


test/basic.js
```diff
@@ -132,7 +132,7 @@ test('publish empty topic throws error', function (t) {
 })
 
 // Catch invalid packet writeToStream errors
-test('catch write errors', function (t) {
+test('return write errors to callback', function (t) {
```
robertsLando, line 135: @mcollina It's covered by this test.

PR #511, opened 2020-06-23
fix: possible leak in handleConnect negate

robertsLando:


lib/handlers/connect.js
```diff
@@ -48,13 +48,10 @@ function handleConnect (client, packet, done) {
...
     if (!err && successful === true) {
       setImmediate(init, client, packet, done)
-      return
-    }
-    client.connecting = false
-    if (err) {
-      client.broker.emit('connectionError', client, err)
+    } else {
+      client.connecting = false
+      done(err)
```
mcollina, line 53: can you add a test?

robertsLando: done

PR #517, opened 2020-07-06
fix: don't pass writeQos errors in packet cb

robertsLando:
deliverQos/deliver0 functions are bounded in subscribe and used then by mqemitters to deliver messages to clients, the callbacks of those functions should NEVER return an error in callback as it is not catched by mqemitter and could cause unhandled exceptions


robertsLando:
@mcollina ping

PR #537, opened 2020-09-03
docs: clusters best practise

robertsLando:


README.md
```diff
@@ -85,6 +86,23 @@ Check Docker docs [here](https://github.com/moscajs/aedes-cli#docker)
...
 
+## Clusters
+
+Aedes needs on disk dbs like MongoDB and Redis in order to work with clusters. Based on our tests and users reports the best performances/stability are reached when using [aedes-persistence-mongodb] paired with [mqemitter-redis].
+
+Other info:
+
+- [aedes-persistence-redis] have some memory leak. Check [this issue](https://github.com/moscajs/aedes-persistence-redis/issues/73#issuecomment-686306508)
+- [mqemitter-mongodb] could have leaks when used in clusters. Check [this issue](https://github.com/mcollina/mqemitter-mongodb/issues/24)
+- You should assign a __FIXED BROKER ID__ to your Aedes instances:
```
mcollina, line 97: This should not be needed.

README.md
```diff
@@ -85,6 +86,23 @@ Check Docker docs [here](https://github.com/moscajs/aedes-cli#docker)
 
 - [Examples](./docs/Examples.md)
 
+## Clusters
+
+Aedes needs on disk dbs like MongoDB and Redis in order to work with clusters. Based on our tests and users reports the best performances/stability are reached when using [aedes-persistence-mongodb] paired with [mqemitter-redis].
+
+Other info:
+
+- [aedes-persistence-redis] have some memory leak. Check [this issue](https://github.com/moscajs/aedes-persistence-redis/issues/73#issuecomment-686306508)
```
mcollina, line 95: I do not think so, as we still do not have a full repro for that issue. A misuse of QoS 1 features can easily look like a memory leak, and we still do not have a repro.

README.md
```diff
@@ -85,6 +86,23 @@ Check Docker docs [here](https://github.com/moscajs/aedes-cli#docker)
...
+- [mqemitter-mongodb] could have leaks when used in clusters. Check [this issue](https://github.com/mcollina/mqemitter-mongodb/issues/24)
```
mcollina, line 96: the mongodb support has a leak, but I think it would be sorted with your fix.

README.md
```diff
@@ -85,6 +86,23 @@ Check Docker docs [here](https://github.com/moscajs/aedes-cli#docker)
...
+- You should assign a __FIXED BROKER ID__ to your Aedes instances:
+
+  ```js
+  var broker = aedes({ id: 'broker' + clusterId })
+  ```
+
+  When a cluster instance dies the new one should take the same id, this is needed because some documents stored in dbs use `brokerId` to identify packets/clients.
+- The repo [aedes-tests](https://github.com/moscajs/aedes-tests) is used to test aedes with clusters and different emitters/persistences. Check its source code to have a starting point on how to work with clusters
```
mcollina, line 104: Overall I would add a few notes that if there are bugs/leaks in production scenarios, we encourage people to send pull request and/or reach out to the maintainer for some paid support.

README.md
robertsLando: Actually persistences store docs using also brokerId as reference, are you sure about this? 

Redis example:
https://github.com/moscajs/aedes-persistence-redis/blob/master/persistence.js#L294
https://github.com/moscajs/aedes-persistence-redis/blob/master/persistence.js#L295
https://github.com/moscajs/aedes-persistence-redis/blob/master/persistence.js#L327
https://github.com/moscajs/aedes-persistence-redis/blob/master/persistence.js#L345

What will happen if this instance dies and a new one comes up with a different id? Will be there any leak?

README.md
robertsLando: Still not 100% sure about this

README.md
mcollina: In theory no, but maybe there are bugs. It's also more expensive so better send the message. There is the whole heartbeath logic in aedes to deal with that.

robertsLando: Removed

mcollina (approved):
lgtm

PR #541, opened 2020-09-14
feat: add $SYS/<brokerId>/new /subscribes and $SYS/<brokerId>/new /unsubscribes #540

### LINKED ISSUE #540: [feat] Could it support client subscribe/unsubscribe notification to other client?
I used mosca.js before, which provide the topic "$SYS/+/new/subscribes"  "$SYS/+/new/unsubscribes"   to let my service client knows that the app client had subscribed the particular topic, but in aedes  it did not publish this message with similar topic.

robertsLando:
Fixes #540

### DIFF
--- a/aedes.js
+++ b/aedes.js
@@ -12,6 +12,7 @@ const Packet = require('aedes-packet')
 const memory = require('aedes-persistence')
 const mqemitter = require('mqemitter')
 const Client = require('./lib/client')
+const { noop } = require('./lib/utils')
 
 module.exports = Aedes.Server = Aedes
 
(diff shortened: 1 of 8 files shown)


aedes.js
```diff
@@ -347,5 +348,3 @@ function PublishState (broker, client, packet) {
   this.client = client
   this.packet = packet
 }
-
-function noop () {}
```
mcollina, line 351: The reason to avoid having a noop in a seperate file is the change of closure context. Please moving it back here.

mcollina (approved):
lgtm

PR #547, opened 2020-10-06
feat: add `packet` to `aedes.preConnect`

robertsLando:
With this feature users will be able to access and modify a connect packet in `aedes.preConnect` handler

This will be a __BREAKING CHANGE__ for users that are using `preConnect`

### DIFF
--- a/aedes.d.ts
+++ b/aedes.d.ts
@@ -4,7 +4,7 @@
 
 /// <reference types="node" />
 
-import { IConnackPacket, IPingreqPacket, IPublishPacket, IPubrelPacket, ISubscribePacket, ISubscription, IUnsubscribePacket } from 'mqtt-packet'
+import { IConnackPacket, IConnectPacket, IPingreqPacket, IPublishPacket, IPubrelPacket, ISubscribePacket, ISubscription, IUnsubscribePacket } from 'mqtt-packet'
 import { AedesPacket } from 'aedes-packet'
 import { Duplex } from 'stream'
 import { Socket } from 'net'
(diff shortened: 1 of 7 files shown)


mcollina (approved):
lgtm

demian85:
Looks really good to me because that's the feature I was looking for. When will this be released?

robertsLando:
@demian85 I created this on purpose after our discussion, I'm going to release this now

robertsLando:
Aedes 0.43.0 is out now

gnought:
@robertsLando Hi, is there something I miss here?
From the initial design, I struggle if we should expose `packet` in `preConnect`. It allows users to modify mqtt packets and will affect all subsequence operations if they did wrongly.
What are the use cases?

robertsLando:
@gnought I don't remember the use case of this but there was a discussion and adding the packet was the only way to allow users to do such thing. Also if someone edits it he should know what he does...

gnought:
@robertsLando could we deep clone the Packet for `preConnect` ?
@mcollina any good approach to fast deep clone an object?

robertsLando:
I think one point of the discussion was the ability to change the connect packet, like force setting the clean flag to true/false on connect

mcollina:
https://www.npmjs.com/package/rfdc is a good bet.

gnought:
Then we need to validate the `CONNECT` packet before going in the `init` function. There is a strong feeling we open for a security issue to mess things up. @robertsLando

gnought:
Or in the other way, we add a hook after authenticate handler and only allow users to modify clean flag, LWT and keep alive on the fly. Sounds better? @robertsLando

robertsLando:
> Or in the other way, we add a hook after authenticate handler and only allow users to modify clean flag, LWT and keep alive on the fly. Sounds better? 

Yep that could be good!

PR #549, opened 2020-10-14
feat: delegate protocol decoding in server-factory and protocol-decoder

getlarge:
Following discussions #413 and [here](https://github.com/moscajs/aedes-server-factory/pull/2).
I removed functions, dependencies, tests, examples and docs to take place in `aedes-protocol-decoder` and in the future `aedes-server-factory` modules.

@robertsLando proposed to assigned `connDetails` in the `client.req`, those details would be extracted during the `createServer` process in `aedes-server-factory`.

In `nextBatch` the buffer would only be a raw mqtt buffer, so no need for the nested conditions anymore.


robertsLando (commented):
@getlarge Could you add a reference to `aedes-server-factory` in readme? Or we could leave that for another PR

robertsLando (approved):
LGTM

robertsLando:
@getlarge Feel free to add yourself to the collaborators section too in readme :)

getlarge:
@robertsLando Ok i will. 
Is it ok to wait that the PRs for server-factory and protocol-decoder will be merged before doing the same for this one ?
We could then add some working examples of the server-factory in this repo.

robertsLando:
Yes I think we need to land: 

1. https://github.com/moscajs/aedes-server-factory/pull/2
2. https://github.com/moscajs/aedes-protocol-decoder/pull/5

And then this one :)

robertsLando:
@getlarge When we are ready with server factory we could merge this first so you can use a fixed version in package.json instead of using the github version

getlarge:
@robertsLando I think server-factory is ready, we would also need a fixed version for the protocol-decoder (which contains breaking changes), which is also a dependency is server-factory.

robertsLando:
@getlarge Ok so we can firstly merge this one, then the protocol decoder, and as last the server factory, right?

robertsLando:
@getlarge this would need some examples docs here also but could do it in a separete pr

getlarge:
@robertsLando I added an example in `Examples.md` (which needs to be corrected), it's not enough?
What kind of example do you have in mind ?

robertsLando (approved):
@getlarge It's ok for me. LGTM.

robertsLando:
@mcollina Can we go on here or would you like to give it a look to?

mcollina (approved):
lgtm

ralphtheninja:
Doesn't this mean a new major version?

robertsLando:
@ralphtheninja AFAIK 0.x version are always like majors (0.x is considered like "initial development" of a project) :) until we don't release a 1.0.0. 

The reason is that many things could change with Mqtt 5 support so we are waiting to see how it goes before releasing 1.0.0

ralphtheninja:
> The reason is that many things could change with Mqtt 5 support so we are waiting to see how it goes before releasing 1.0.0

Ok! Just checking :+1:

PR #563, opened 2020-11-18
Fewer Last Will calls when last will doesn't change between sessions

mogupta:


robertsLando (changes requested):
All this changes would break actual will management in a cluster env

mogupta:
Thanks @robertsLando for the feedback. 
Can you elaborate a bit to point me in the right direction ? I tested on the cluster enviroment briefly , will seems to work without problem. i.e when device connects and when it goes offline, the broker do emit the last will.

mcollina:
I'm pretty sure that landing this would cause a data leak in the database as well as a MQTT spec violation :(.

PR #567, opened 2020-12-07
fix: replace shortid with hyperid #565

### LINKED ISSUE #565: [question] Conflict with Passenger and Node Cluster module
Aedes uses shortid 
But shortid required the Node Cluster module, which is incompatible with Passenger, a non-functional shim was returned and your app may still work. 
However, please remove or [replace](https://github.com/ai/nanoid/) the related package as soon as possible.

robertsLando:
Fixes #565

### DIFF
--- a/docs/Client.md
+++ b/docs/Client.md
@@ -62,7 +62,7 @@ a read-only flag indicates if client is closed or not.
 
 ## client.id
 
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

PR #583, opened 2021-01-29
fix: `emptyQueue` hitting its high water mark

scagood:
This is the same issue as reported here:
https://github.com/rvagg/through2/issues/32#issuecomment-62645342

This orignial filled the streams internal buffer because there is no destination stream.
The "client" object returned as the second argument in `writeCallback` was trying to be forwarded

### DIFF
--- a/test/qos1.js
+++ b/test/qos1.js
@@ -509,6 +509,71 @@ test('resend publish on non-clean reconnect QoS 1', function (t) {
   })
 })
 
+test('resend many publish on non-clean reconnect QoS 1', function (t) {
+  t.plan(38)
+
+  const broker = aedes()
+  t.tearDown(broker.close.bind(broker))
+
+  const opts = { clean: false, clientId: 'abcde' }
+  let subscriber = connect(setup(broker), opts)
+
+  subscribe(t, subscriber, 'hello', 1, function () {
+    subscriber.inStream.end()
+
+    const publisher = connect(setup(broker))
+
+    let count = 0
+    while (++count <= 17) {
+      publisher.inStream.write({
+        cmd: 'publish',
+        topic: 'hello',
+        payload: 'message-' + count,
+        qos: 1,
+        messageId: 42 + count
+      })
+    }
+    publisher.outStream.once('data', function (packet) {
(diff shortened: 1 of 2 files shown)


robertsLando:
@scagood Could you add a test for this?

robertsLando:
cc @mcollina could you give a look at this?

mcollina:
Might make sense... but I need to see a test to repro this.

scagood:
Hi, sorry I didnt add a test earlier (that's why the PR is a draft :sweat_smile:)

After far to long struggling with tap I think I added a reasonable test. Do tell me if its not so good I can fix it!

test/qos1.js
```diff
@@ -509,6 +509,71 @@ test('resend publish on non-clean reconnect QoS 1', function (t) {
...
+  const opts = { clean: false, clientId: 'abcde' }
+  let subscriber = connect(setup(broker), opts)
+
+  subscribe(t, subscriber, 'hello', 1, function () {
+    subscriber.inStream.end()
+
+    const publisher = connect(setup(broker))
+
+    let count = 0
+    while (++count <= 17) {
```
scagood, line 527: The count here needs to be 17 (or higher) to ensure we're above the [`through2` highWaterMark](https://github.com/rvagg/through2/blob/master/through2.js#L70)

mcollina (approved):
lgtm

robertsLando (approved):
Lgtm

robertsLando:
Thanks for the fix @scagood. Could you check if there are also other parts in the code where this could be applied?

scagood:
:thinking: I've just done a quick 'once over' and the only places I can see this being an issue are:
- Client#deliverQoS [lib/client.js:127](https://github.com/moscajs/aedes/blob/master/lib/client.js#L127)
- in persistence plugins in `outgoingClearMessageId` from: [lib/client.js:376](https://github.com/moscajs/aedes/blob/master/lib/client.js#L376) for example [aedes-persistence/persistence.js:233](https://github.com/moscajs/aedes-persistence/blob/master/persistence.js#L233)

robertsLando:
Ok let me know! @scagood

PR #584, opened 2021-02-23
feat(bridge): support incoming mosquitto bridge connections

### LINKED ISSUE #329: [feat] Bridge
Is it possible to bridge an aedes broker to a mosquitto broker?

phil-mitchell:
The incoming mosquitto bridge connections add 128 to the protocol
version to indicate ras=true and nl=true.

Support for this was added in mqtt-packet with
https://github.com/mqttjs/mqtt-packet/pull/106

Subscriptions from these connections will have ras=true and nl=true

When publishing to these connections, aedes will now
copy the Packet and set retain to false only if ras is not true.

Fixes #329

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ Check Docker docs [here](https://github.com/moscajs/aedes-cli#docker)
 - [Dynamic Topics][dynamic_topics] Support
 - MQTT Bridge Support between aedes
 - [MQTT 5.0][mqttv5] _(not support yet)_
-- [Bridge Protocol][bridge_protocol]
+- [Bridge Protocol][bridge_protocol] _(incoming connections only)_
 
 ## Examples
 
--- a/README.md
+++ b/README.md
@@ -95,6 +95,15 @@ Other info:
 
 - The repo [aedes-tests](https://github.com/moscajs/aedes-tests) is used to test aedes with clusters and different emitters/persistences. Check its source code to have a starting point on how to work with clusters
 
+## Bridge connections
+
+Normally, when publishing a message, the `retain` flag is consumed by Aedes and
+then set to `false` so that recipients of the message do not retain it as well.
+
+Brokers that support the [Bridge Protocol][bridge_protocol] can connect to
+Aedes.  When connecting with this special protocol, subscriptions work as usual
+excecpt that the `retain` flag in the packet is propagated as-is.
+
 ## Exensions
 
 - [aedes-logging]: Logging module for Aedes, based on Pino
--- a/README.md
+++ b/README.md
@@ -98,11 +98,16 @@ Other info:
 ## Bridge connections
 
 Normally, when publishing a message, the `retain` flag is consumed by Aedes and
-then set to `false` so that recipients of the message do not retain it as well.
+then set to `false`.  This is done for two reasons:
+
+- MQTT-3.3.1-9 states that it MUST set the RETAIN flag to 0 when a PUBLISH
+  Packet is sent to a Client because it matches an established subscription
+  regardless of how the flag was set in the message it received.
+- When operating as a cluster, only one Aedes node may store the packet
 
 Brokers that support the [Bridge Protocol][bridge_protocol] can connect to
 Aedes.  When connecting with this special protocol, subscriptions work as usual
-excecpt that the `retain` flag in the packet is propagated as-is.
+except that the `retain` flag in the packet is propagated as-is.
 
 ## Exensions
 
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ Check Docker docs [here](https://github.com/moscajs/aedes-cli#docker)
 - [Dynamic Topics][dynamic_topics] Support
 - MQTT Bridge Support between aedes
 - [MQTT 5.0][mqttv5] _(not support yet)_
-- [Bridge Protocol][bridge_protocol] _(not support yet)_
+- [Bridge Protocol][bridge_protocol]
 
 ## Examples
 
--- a/aedes.js
+++ b/aedes.js
@@ -256,46 +256,11 @@ Aedes.prototype.publish = function (packet, client, done) {
   this._series(new PublishState(this, client, packet), publishFuncs, p, done)
 }
 
-function createRetainToFalseWrapper (broker, topic, func) {
-  func._retainToFalseWrappers = func._wrappers || {}
-  func._retainToFalseWrappers[topic] = func._retainToFalseWrappers[topic] || {}
-  func._retainToFalseWrappers[topic][broker] = function handlePacketSubscription (_packet, cb) {
-    _packet = new Packet(_packet, broker)
-    _packet.retain = false
-    func(_packet, cb)
-  }
-  func._retainToFalseWrappers[topic][broker]._freeWrapper = function freeWrapper () {
-    delete ((func._retainToFalseWrappers || {})[topic] || {})[broker]
-    if (Object.keys((func._retainToFalseWrappers || {})[topic] || {}).length === 0) {
-      delete func._retainToFalseWrappers[topic]
-    }
-    if (Object.keys(func._retainToFalseWrappers).length === 0) {
-      delete func._retainToFalseWrappers
-    }
-  }
-  return func._retainToFalseWrappers[topic][broker]
-}
-
-Aedes.prototype.subscribe = function (topic, func, rap, done) {
-  if (done === undefined && typeof rap === 'function') {
-    done = rap
-    rap = false
-  }
-  if (!rap) {
-    func = createRetainToFalseWrapper(this, topic, func)
-  }
+Aedes.prototype.subscribe = function (topic, func, done) {
   this.mq.on(topic, func, done)
 }
 
 Aedes.prototype.unsubscribe = function (topic, func, done) {
-  if (func) {
-    func = ((func._retainToFalseWrappers || {})[topic] || {})[this] || func
-  }
-
-  if (func && func._freeWrapper) {
-    func._freeWrapper()
-  }
-
   this.mq.removeListener(topic, func, done)
 }
 
--- a/aedes.js
+++ b/aedes.js
@@ -256,11 +256,46 @@ Aedes.prototype.publish = function (packet, client, done) {
   this._series(new PublishState(this, client, packet), publishFuncs, p, done)
 }
 
-Aedes.prototype.subscribe = function (topic, func, done) {
+function createRetainToFalseWrapper (broker, topic, func) {
+  func._retainToFalseWrappers = func._wrappers || {}
+  func._retainToFalseWrappers[topic] = func._retainToFalseWrappers[topic] || {}
+  func._retainToFalseWrappers[topic][broker] = function handlePacketSubscription (_packet, cb) {
+    _packet = new Packet(_packet, broker)
+    _packet.retain = false
+    func(_packet, cb)
+  }
+  func._retainToFalseWrappers[topic][broker]._freeWrapper = function freeWrapper () {
+    delete ((func._retainToFalseWrappers || {})[topic] || {})[broker]
+    if (Object.keys((func._retainToFalseWrappers || {})[topic] || {}).length === 0) {
+      delete func._retainToFalseWrappers[topic]
+    }
+    if (Object.keys(func._retainToFalseWrappers).length === 0) {
+      delete func._retainToFalseWrappers
+    }
+  }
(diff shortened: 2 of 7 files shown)


aedes.js
```diff
@@ -170,7 +170,6 @@ function storeRetained (packet, done) {
 }
 
 function emitPacket (packet, done) {
-  packet.retain = false
```
robertsLando, line 173: Why? This is used to prevent other broker instances to store this packet again when it's emiitted

lib/handlers/subscribe.js
```diff
@@ -23,9 +23,12 @@ function SubAck (packet, granted) {
   this.granted = granted
 }
 
-function Subscription (qos, func) {
+function Subscription (qos, func, rh, rap, nl) {
   this.qos = qos
   this.func = func
+  this.rh = rh
+  this.rap = rap
+  this.nl = nl
```
robertsLando, line 31: Could you add a comment to explain this properties?

robertsLando (changes requested):

lib/handlers/subscribe.js
```diff
@@ -119,19 +125,30 @@ function addSubs (sub, done) {
...
+  const nl = this.nl
+  const deliverFunc = qos > 0 ? client.deliverQoS : client.deliver0
+
+  let func = function handlePacketSubscription (_packet, cb) {
+    if (!rap) {
+      _packet = new Packet(_packet, broker)
+      _packet.retain = false
+    }
+    deliverFunc(_packet, cb)
+  }
```
mcollina, line 139: Please avoid to allocate this function if rap is set.

aedes.js
mcollina: Unfortunately this cannot be easily removed.

mcollina (changes requested):

aedes.js
phil-mitchell: We need to keep the retain on the packet until we are able to check if "retain as published" (rap) is true.  That's the whole point of the rap flag.  The bridge protocol enables retain-as-published.  I moved this into the subscription handler.  What are the concerns with moving this into the subscription handler?

lib/handlers/subscribe.js
phil-mitchell: Will do

lib/handlers/subscribe.js
phil-mitchell: Will do.

README.md
```diff
@@ -95,6 +95,15 @@ Other info:
 
 - The repo [aedes-tests](https://github.com/moscajs/aedes-tests) is used to test aedes with clusters and different emitters/persistences. Check its source code to have a starting point on how to work with clusters
 
+## Bridge connections
+
+Normally, when publishing a message, the `retain` flag is consumed by Aedes and
+then set to `false` so that recipients of the message do not retain it as well.
```
robertsLando, line 101: That's not really correct, it is set to false because Aedes can be used in a cluster env, in such cases the mqemitter is used to share a message between different broker instances, the retain is set to false to prevent other instances to store it again

robertsLando: @mcollina

aedes.js
mcollina: can you link it to the block of code that replaces this?

phil-mitchell: @mcollina this is what handlePacketSubscription in subscribe.js is doing.  Instead of globally setting it to false for all subscriptions, it sets it to false only if the retain-as-published flag is not set on the subscription.

If clusters go through a different (untested?) path, then clearing the retain flag can happen on that path, or I could set up a different emitter for client subscriptions so that clusters always get the flag set to false while bridges and regular clients respect the rap flag.

robertsLando: I think as an alternative we could add a `_stored` property to the packet to prevent others to store it again?

README.md
mcollina: @robertsLando that's correct.

aedes.js
mcollina: I think we should verify if this is still needed, e.g. if the packet gets stored multiple times when using a cluster.

robertsLando: This only happens when a user sends a packet using `broker.publish` function. BTW I think that it could be safely removed by double checking the code, the 'worst' thing that could happen is some useless write to the db

README.md
phil-mitchell: Probably also true, but my understanding of the MQTT spec is that the `retain` flag needs to be false when received through a subscribe.

http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html

> It MUST set the RETAIN flag to 0 when a PUBLISH Packet is sent to a Client because it matches an established subscription regardless of how the flag was set in the message it received [MQTT-3.3.1-9].

aedes.js
phil-mitchell: I had thought of the `_stored` idea, but wasn't sure whether it would cause other issues.  In any case, it becomes the event listener's responsibility to not retain the packet if it's not supposed to.

As I've said, the tests pass, including some new ones I added, so how can I verify this concern?

README.md
robertsLando: First of all I would add this comment there, then sincerly I dunno what's the best way to continue as both seems to have a reson to be accepted. I see no reference on bridge retain handling on specs

phil-mitchell: Bridge retain handling is outside of the standard, so the only spec about bridge retain handling is here: https://github.com/mqtt/mqtt.github.io/wiki/bridge_protocol

However, MQTT v5 does mention retain handling: https://docs.oasis-open.org/mqtt/mqtt/v5.0/mqtt-v5.0.html

> The setting of the RETAIN flag in an Application Message forwarded by the Server from an established connection is controlled by the Retain As Published subscription option. Refer to section 3.8.3.1 for a definition of the Subscription Options.
> ·         If the value of Retain As Published subscription option is set to 0, the Server MUST set the RETAIN flag to 0 when forwarding an Application Message regardless of how the RETAIN flag was set in the received PUBLISH packet [MQTT-3.3.1-12].
> ·         If the value of Retain As Published subscription option is set to 1, the Server MUST set the RETAIN flag equal to the RETAIN flag in the received PUBLISH packet [MQTT-3.3.1-13].

aedes.js
robertsLando: > As I've said, the tests pass, including some new ones I added, so how can I verify this concern?

You should add a test where you do a broker subscribe followed by a broker publish to check if the received message has the retained flag cleared or not

phil-mitchell: I added the test and, indeed, the retain flag was previously set to false and is now true.  I've pushed a change with a proposed solution.  The publish function now takes the `rap` parameter and applies the wrapper function that I had previously placed in `subscription.js`.  This way all callers of `subscribe` have the retain-as-published functionality available to them but default to the standard behaviour of setting `retain` to `false`.

mcollina (changes requested):
I think you can move the wrapper inside lib/handlers/subscribe instead.

phil-mitchell:
> I think you can move the wrapper inside lib/handlers/subscribe instead.

To be clear, you'd like me to put it back the way it was, knowing the behaviour change to those who subscribe directly to the broker?

mcollina:
> To be clear, you'd like me to put it back the way it was, knowing the behaviour change to those who subscribe directly to the broker?

No, I didn't really understand why the function was moved out of subscribe.

phil-mitchell:
> > To be clear, you'd like me to put it back the way it was, knowing the behaviour change to those who subscribe directly to the broker?
> 
> No, I didn't really understand why the function was moved out of subscribe.

Okay, I misunderstood your comment.  Yes, I'll do that.  It was mostly moved out to avoid excessive nesting (i.e. for readability).

phil-mitchell:
> > To be clear, you'd like me to put it back the way it was, knowing the behaviour change to those who subscribe directly to the broker?
> 
> No, I didn't really understand why the function was moved out of subscribe.

It was moved out because the broker subscribe doesn't go through lib/handlers/subscribe.  So it was moved to share the functionality with broker subscribe.

mcollina:
Why does the broker subscribe need it?

phil-mitchell:
> Why does the broker subscribe need it?

@robertsLando said so?

phil-mitchell:
> Why does the broker subscribe need it?

If broker.subscribe doesn't add this wrapper, then when I remove the retain = false from emitPacket the subscribers who call broker.subscribe have retain = true.  There has been a long conversation with you and @robertsLando saying that would be a bad thing.

phil-mitchell:
Ultimately I don't care, but if you can give me consistent direction, I'll do it.  Otherwise, I'll just use my fork of this repo the way it works for me.

mcollina:
> > Why does the broker subscribe need it?

> @robertsLando said so?

 Multiple maintainers can have different view about the project, that's the good/bad side of having open governance.

To recap, I would recommend:

* no modification in `emitPacket`
* no modification in `broker.subscribe`

WDYT @robertsLando?

What will leave us with then? I'm sorry but you are probably more into this feature than me at this point - I'd need some guidance.

robertsLando:
> To recap, I would recommend:
> no modification in emitPacket
> no modification in broker.subscribe

I agree with @mcollina expecially latest commit has introduced some complexity that makes the code hard to read. I think i may have explained wrong in my [comment](https://github.com/moscajs/aedes/pull/584#discussion_r582823074), my idea was to create a strict test for the `packet.retain = false` that you removed (as it seems was not covered as the tests works with and without it). I would revert [`b4ad77d` (#584)](https://github.com/moscajs/aedes/pull/584/commits/b4ad77dfa7491b5f64e67b9266670a90ed4073a1) except for the changes on readme that are ok

phil-mitchell:
Okay, so I guess I misinterpreted your request to add that test.  You wanted me to test the new behaviour where retain is *not* set to false in broker.subscribe.  I have reverted to the behaviour where retain is *not* modified in emitPacket.

mcollina:
I think we keep misinterpreting ourselves (maybe it's the language).

* no code changes in emitPacket, retain = false should be kept there.
* no code changes in subscribe, there should be no automatic wrapping.

Is this doable while supporting the protocol?

phil-mitchell:
> I think we keep misinterpreting ourselves (maybe it's the language).
> 
> * no code changes in emitPacket, retain = false should be kept there.
> * no code changes in subscribe, there should be no automatic wrapping.
> 
> Is this doable while supporting the protocol?

If retain is false when it reaches subscriber.js (which subscribes to the event via broker.subscribe), then it is not possible to know what its value *was* in order to set it back to true.  If we introduce another field in the Packet to indicate that it *was* retained, then we could use that, but that might have other side-effects.

robertsLando (approved):

lib/handlers/subscribe.js
```diff
@@ -119,19 +136,31 @@ function addSubs (sub, done) {
   const broker = client.broker
   const topic = sub.topic
   const qos = sub.qos
+  const rh = this.rh
+  const rap = this.rap
+  const nl = this.nl
   let func = qos > 0 ? client.deliverQoS : client.deliver0
 
+  if (!rap) {
```
robertsLando, line 144: @phil-mitchell can rap be undefined too or is it true by default? Would a check `rap === false` be better?

robertsLando (changes requested):
Sorry pressed the approve changes by error... From a code side it looks almost all good to me, @phil-mitchell I think that @mcollina would prefer to don't create the wrapping function in the `doSubscribe` function but to clear the retained flag in the client `deliverQoS` and `deliver0` functions instead. This is the only missing thing right now I think?

lib/handlers/subscribe.js
phil-mitchell: @robertsLando rap  can be undefined, in which case it should be considered false.  i.e. rap = false is default.  So any falsey value of rap should set `retain = false`

phil-mitchell:
> Sorry pressed the approve changes by error... From a code side it looks almost all good to me, @phil-mitchell I think that @mcollina would prefer to don't create the wrapping function in the `doSubscribe` function but to clear the retained flag in the client `deliverQoS` and `deliver0` functions instead. This is the only missing thing right now I think?

I don't see where @mcollina requested that.  Sorry, but I don't see how this is feasible unless we pass another parameter to these methods.  The subscription coming out of `mqtt-packet` is the thing that knows about the `rap` flag, not the client.

mcollina (approved):
LGTM

I thought a bit more about this and I think this code is ok.

PR #603, opened 2021-03-30
fix: set the dup flag for QoS 2

tomekt-bolt:
This PR adds setting the `dup` flag for re-deliveries of QoS 2 packets.

It requires [complementary changes](https://github.com/tomekt-bolt/aedes-persistence/commit/eeb85d94226ed31eefaf3fad26c5bf3179bdfb7a) in the `aedes-persistence` project and possibly some adjustments in concrete persistence layer implementations.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -124,10 +124,13 @@ function Client (broker, conn, req) {
         if (clientSub && (clientSub.qos || 0) < packet.qos) {
           packet.qos = clientSub.qos
         }
-        packet.writeCallback = cb
         if (that.clean || packet.retain) {
+          packet.writeCallback = cb
           writeQoS(null, that, packet)
         } else {
+          packet.writeCallback = () => {
+            packet.setDupIfNecessary(broker.persistence, that, cb)
+          }
           broker.persistence.outgoingUpdate(that, packet, writeQoS)
         }
       })
--- a/lib/client.js
+++ b/lib/client.js
@@ -129,6 +129,7 @@ function Client (broker, conn, req) {
           writeQoS(null, that, packet)
         } else {
           packet.writeCallback = () => {
+            // [MQTT-3.3.1-1]
             packet.setDupIfNecessary(broker.persistence, that, cb)
           }
           broker.persistence.outgoingUpdate(that, packet, writeQoS)
--- a/lib/client.js
+++ b/lib/client.js
@@ -6,7 +6,7 @@ const util = require('util')
 const eos = require('end-of-stream')
 const Packet = require('aedes-packet')
 const write = require('./write')
-const QoSPacket = require('./qos-packet')
+const { QoSPacket, setDupIfNecessary } = require('./qos-packet')
 const handleSubscribe = require('./handlers/subscribe')
 const handleUnsubscribe = require('./handlers/unsubscribe')
 const handle = require('./handlers')
@@ -130,7 +130,7 @@ function Client (broker, conn, req) {
         } else {
           packet.writeCallback = () => {
             // [MQTT-3.3.1-1]
-            packet.setDupIfNecessary(broker.persistence, that, cb)
+            setDupIfNecessary(packet, broker.persistence, that, cb)
           }
           broker.persistence.outgoingUpdate(that, packet, writeQoS)
         }
--- a/lib/qos-packet.js
+++ b/lib/qos-packet.js
@@ -18,6 +18,14 @@ function QoSPacket (original, client) {
   } else {
     this.messageId = original.messageId
   }
+
+  this.setDupIfNecessary = (persistence, client, cb) => {
+    if (this.qos === 2) {
+      persistence.outgoingUpdate(client, { ...this, dup: true }, cb)
+    } else {
+      cb()
+    }
+  }
 }
 
 util.inherits(QoSPacket, Packet)
--- a/lib/qos-packet.js
+++ b/lib/qos-packet.js
@@ -18,16 +18,19 @@ function QoSPacket (original, client) {
   } else {
     this.messageId = original.messageId
   }
+}
 
-  this.setDupIfNecessary = (persistence, client, cb) => {
-    if (this.qos === 2) {
-      persistence.outgoingUpdate(client, { ...this, dup: true }, cb)
-    } else {
-      cb()
-    }
+function setDupIfNecessary (packet, persistence, client, cb) {
+  if (packet.qos === 2) {
+    persistence.outgoingUpdate(client, { ...packet, dup: true }, cb)
+  } else {
+    cb()
   }
 }
 
 util.inherits(QoSPacket, Packet)
 
-module.exports = QoSPacket
+module.exports = {
+  QoSPacket,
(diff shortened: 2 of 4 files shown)


robertsLando:
@tomekt-bolt There is a test that is failing. COuld you also attach links to mqtt specs on the peaces of code using dup? For example: http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc385349262 ? Capther [MQTT-3.3.1-1]

lib/client.js
```diff
@@ -124,10 +124,13 @@ function Client (broker, conn, req) {
...
         }
-        packet.writeCallback = cb
         if (that.clean || packet.retain) {
+          packet.writeCallback = cb
           writeQoS(null, that, packet)
         } else {
+          packet.writeCallback = () => {
+            packet.setDupIfNecessary(broker.persistence, that, cb)
+          }
           broker.persistence.outgoingUpdate(that, packet, writeQoS)
```
robertsLando, line 134: This will trigger two `outgoingUpdate` one just after the other, I think we could do this in one

lib/client.js
```diff
@@ -124,10 +124,13 @@ function Client (broker, conn, req) {
...
           packet.qos = clientSub.qos
         }
-        packet.writeCallback = cb
         if (that.clean || packet.retain) {
+          packet.writeCallback = cb
           writeQoS(null, that, packet)
         } else {
+          packet.writeCallback = () => {
+            packet.setDupIfNecessary(broker.persistence, that, cb)
+          }
```
robertsLando, line 133: This will trigger two `outgoingUpdate` one just after the other, I think we could do this in only one instead

robertsLando (changes requested):

robertsLando:
Also this requires updates on mongodb and redis persistence (at least):

https://github.com/moscajs/aedes-persistence-mongodb/blob/67ca807e38ef647340ab3bc63b6a1050185cf97f/persistence.js#L507

https://github.com/moscajs/aedes-persistence-redis/blob/master/persistence.js#L412

lib/client.js
tomekt-bolt: I'm not sure about it. Isn't the order of the operations crucial here? Like:
1. Do the original update that was there before my changes.
2. Actually write the message to the stream.
3. Update the `dup` flag.

If something goes wrong in point 2. - eg. the stream is no longer available, then I suppose we don't want the `dup` flag updated.

robertsLando: yeah you right by looking at this again seems there are no alternatives. From a code side it seems ok to me I would also like to know @mcollina and @gnought thoughts

gnought:
I'm against the PR. We should add a dup flag in aedes packet.

robertsLando:
aedes packet already has a dup flag: https://github.com/moscajs/aedes-packet/blob/master/packet.js#L11

tomekt-bolt:
> Also this requires updates on mongodb and redis persistence (at least)

`aedes-persistence-redis` works out of the box, as it updates the stored value with a freshly encoded packet: https://github.com/moscajs/aedes-persistence-redis/blob/9cc9864e512059ebb3add14ce7463544ad12b7fe/persistence.js#L335.

As for mongodb, I'd require some help (provided we're happy with the general direction of the change), as I have no competences in this area.

tomekt-bolt:
> `aedes-persistence-redis` works out of the box, as it updates the stored value with a freshly encoded packet: https://github.com/moscajs/aedes-persistence-redis/blob/9cc9864e512059ebb3add14ce7463544ad12b7fe/persistence.js#L335.

OK, now I see that it probably _doesn't_ work out of the box, as the packet updated in the line I've pointed is shared among all the interested clients, so we're setting the `dup` flag for all of them here. This means, that the change required for `aedes-prersistence-redis` is not that trivial, as we need to store the `dup` flag on per-client basis.

Is this correct?

mnowotka:
Hey @robertsLando! Will you move forward with this PR? 

In the aedes readme it says:

```
Full compatible with MQTT 3.1 and 3.1.1
```

But without setting the dup flag I'm afraid this is not entirely true for the case of QoS == 2.
If this change is not acceptable, can you please explain how can we contribute to fix this issue?

mnowotka:
And I forgot to say: Many thanks for writing aedes, this is a great piece of software, thanks for making it open!

robertsLando:
In order to move on wih this we firstly need to fix persistences. I would also know why @gnought is against this and how he think would be a better implementation of it, and also @mcollina thoughts about this as he is the main author of Aedes and his understanding in mqtt and Aedes is much better then mine

tomekt-bolt:
Thanks @robertsLando for your comment. I will wait for some more feedback on the general direction of this change before investing in aligning the persistence layer implementations then.

robertsLando:
@tomekt-bolt Could you link opened PR on persisistences please?

tomekt-bolt:
I don't have any open PRs on persistences related to this issue. As I've mentioned above, I'd rather not invest in aligning the persistence without having a green light for the general approach of this PR.

I understand that the persistence changes would need to be applied _prior_ to this PR, but what I want to avoid is spending the time on persistence PRs and later finding out that there's some inherent problem with what I've proposed here.

robertsLando:
Agree, let's see what @gnought and @mcollina think about this so

lib/qos-packet.js
```diff
@@ -18,6 +18,14 @@ function QoSPacket (original, client) {
...
     this.messageId = original.messageId
   }
+
+  this.setDupIfNecessary = (persistence, client, cb) => {
+    if (this.qos === 2) {
+      persistence.outgoingUpdate(client, { ...this, dup: true }, cb)
+    } else {
+      cb()
+    }
+  }
```
mcollina, line 28: Move this to a top level function, allocating one for each packet is too expensive.

mcollina (changes requested):

lib/qos-packet.js
tomekt-bolt: Done.

tomekt-bolt:
Hi @robertsLando , @gnought, @mcollina,

thanks for your feedback so far. Could you please let me know where do we stand with this subject at the moment?

Do you intend to merge this PR, provided the required changes would be added to the persistence implementations? If so, can I count on some assistance in updating the persistence parts?

Or is there some inherent problem with my approach?

mcollina (approved):
lgtm

mcollina:
a test is failing

tomekt-bolt:
> a test is failing

The failing tests are the ones I've added, which require the [mentioned change in aedes-persistence](https://github.com/tomekt-bolt/aedes-persistence/commit/eeb85d94226ed31eefaf3fad26c5bf3179bdfb7a), I can open another PR for that change, but don't know how to link it to be used for tests in this one.
https://github.com/moscajs/aedes/pull/603#issuecomment-815806689

robertsLando:
> I can open another PR for that change, but don't know how to link it to be used for tests in this one.

The procedure is to firstly fix and merge the changes needed, update the package here and then merge this one

PR #605, opened 2021-03-31
Added a preUnsubscribe handler, analog to authorizeSubscribe

chfritz:
Invoked then a client unsubscribes from topics. Just like authorizeSubscribe, this function allows modifying the topics because executing the unsubscribe. This is needed in cases where topics are modified in authorizeSubscribe. For example:

```js
aedes.authorizeSubscribe = (client, subscription, callback) => {
  // overwrite subscription: force client to its namespace
  subscription.topic = `/${client.id}/${subscription.topic}`;
  callback(null, subscription);
}

aedes.preUnsubscribe = (client, packet) => {
  // overwrite unsubscriptions: force client to its namespace
  for (let i in packet.unsubscriptions) {
    packet.unsubscriptions[i] = `/${client.id}/${packet.unsubscriptions[i]}`;
  }
}
```


docs/Aedes.md
```diff
@@ -386,6 +387,28 @@ aedes.authorizeForward = function (client, packet) {
 }
 ```
 
+## Handler: preUnsubscribe (client, packet)
+
+- client: [`<Client>`](./Client.md)
+- packet: `<aedes-packet>` & [`UNSUBSCRIBE`][UNSUBSCRIBE]
+
+Invoked then a client unsubscribes from topics. Just like authorizeSubscribe, this function allows modifying the topics before executing the unsubscribe. This may be needed in cases where topics were modified in authorizeSubscribe. For example:
```
robertsLando, line 395:
```suggestion
Invoked when a client unsubscribes from topics. Just like `authorizeSubscribe`, this function allows modifying the topics before executing the unsubscribe. This may be needed in cases where topics were modified in authorizeSubscribe. For example:
```

robertsLando (changes requested):

aedes.js
```diff
@@ -69,6 +69,7 @@ function Aedes (opts) {
   this.authorizePublish = opts.authorizePublish
   this.authorizeSubscribe = opts.authorizeSubscribe
   this.authorizeForward = opts.authorizeForward
+  this.preUnsubscribe = opts.preUnsubscribe
```
robertsLando, line 72: You should add a `defaultPreSubscribe` function and add it to `defaultOptions`

robertsLando (changes requested):

aedes.js
chfritz: Ah, yes, good point. Thanks, done.

robertsLando: Should that allow some async stuff like the others do? maybe adding a callbackk to it?

robertsLando (changes requested):

gnought:
The main question is what's the use case of modifying subscription topic in `authorizeSubscribe` ?

robertsLando:
@gnought You mean `preSubscribe` ?

chfritz:
@gnought the example I'm providing in the description is the use-case I need it for: I need to force clients into their own namespace, without the clients needing to know about their namespaces. Clients may subscribe to `#`, i.e., get everything, but will actually be reduced to `/${client.id}/#`, i.e., everything in their namespace only.

chfritz:
@robertsLando I've made `preUnsubscribe` async as you suggested.

docs/Aedes.md
```diff
@@ -29,6 +29,7 @@
   - [Handler: authorizePublish (client, packet, callback)](#handler-authorizepublish-client-packet-callback)
   - [Handler: authorizeSubscribe (client, subscription, callback)](#handler-authorizesubscribe-client-subscription-callback)
   - [Handler: authorizeForward (client, packet)](#handler-authorizeforward-client-packet)
+  - [Handler: preUnsubscribe (client, packet)](#handler-preunsubscribe-client-packet)
```
robertsLando, line 32: remember to update signature here and later, callback is missing

robertsLando (changes requested):

gnought:
I believe you also hook `authorizePublish` to modify the PUBLISH topic, am I right?

chfritz:
yes, exactly, @gnought .

mcollina (changes requested):
Can you please add a unit test?

gnought:
How about if we add the support of client isolation? Will it fit in your case? @chfritz

chfritz:
@mcollina : will do.
@gnought : maybe. What do you mean by support for client isolation?

gnought:
Just like your use case, we prepend some client-specific string in client topics.

chfritz:
I see. Yes, I think that would work for me.

PR #625, opened 2021-04-22
fix: upgrade Aedes.d.ts and Client to extends from EventEmitter

cjimenezsaiz:
Aedes and Client interfaces has been upgraded as an extends from EventEmitter in order to allow a complete events management in TypeScript applications


robertsLando:
Please fix tests

```
types/client.d.ts:4:7
  ✖  4:7  Module "events" can only be default-imported using the esModuleInterop flag  

  types/instance.d.ts:5:7
  ✖  5:7  Module "events" can only be default-imported using the esModuleInterop flag  
```

robertsLando (approved):

PR #596, opened 2021-03-14
Use eslint, and refactor typings

gnought:
1. Upgrade to eslint
2. Modularize typings into types/ directory
3. Fix `ConnectPacket` in `preConnect` handler
4. Fix typing test to cover missing broker events, functions and some properties

PR #630, opened 2021-04-28
fix/typings: Remove Readonly<T> from AuthenticateHandler params

bdchauvette:
This commit removes the Readonly<T> wrappings around the `username` and `password` params to the `AuthenticateHandler` interface, which were added in #596.

The motivation for this change is that when `password` is `Readonly<Buffer>`, the type is incompatible with `crypto.timingSafeEqual`, which is the function Aedes users should be using to compare raw, sensitive buffers with each other.
Because `Readonly<Buffer>` is incompatible with `crypto.timingSafeEqual`, users end up having to cast with `password as Buffer`, which largely defeats the purpose of marking it `Readonly` in the first place and introduces casting in security-related areas of the code where it's not really needed in the first place.

The error it gives is:

    No overload matches this call.
      Overload 1 of 2, '(a: ArrayBufferView, b: ArrayBufferView): boolean', gave the following error.
        Argument of type 'Readonly<Buffer>' is not assignable to parameter of type 'ArrayBufferView'.
          Type 'Readonly<Buffer>' is missing the following properties from type 'Float32Array': [Symbol.iterator], [Symbol.toStringTag]
      Overload 2 of 2, '(a: ArrayBufferView, b: ArrayBufferView): boolean', gave the following error.
        Argument of type 'Readonly<Buffer>' is not assignable to parameter of type 'ArrayBufferView'.ts(2769)

Removing it from `username` has no effect, because strings are already immutable in JavaScript, and TypeScript will automatically treat it as if it were just `string`.

### DIFF
--- a/test/types/aedes.test-d.ts
+++ b/test/types/aedes.test-d.ts
@@ -1,4 +1,5 @@
 import { expectType } from 'tsd'
+import { timingSafeEqual } from 'crypto'
 import { Socket } from 'net'
 import type {
   Aedes,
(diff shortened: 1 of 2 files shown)


robertsLando (commented):
LGTM

robertsLando (approved):

getlarge (approved):
Even though i understand the intent of marking them as Readonly, i also had the same typing problem.

getlarge:
I think @gnought introduced this type change few weeks ago, it would be nice to have his opinion before merging.

robertsLando:
@mcollina Thoughts?

Issue #147, opened 2017-08-07
Username and password are not provided to some auth methods


martindederer:
Currently the username and password are not accessible from within the custom authorizePublish and authorizeSubscribe methods.

This becomes an issue when authenticating and authorizing with JWT or similar token based auth schemes is desired. In this case the decision to allow or deny a request to publish or subscribe to a certain MQTT topic depends on the payload of the auth token. The widespread solution for using auth tokens with the MQTT protocol seems to be to send the auth token in the username or password field on connection initialization.

Would you be interested in a PR that stores the username and password in the client instance on connection initialization? I would be happy to provide it in that case.

mcollina:
@martindederer you can store them in the client object in `authenticate`. You can also add all the state you might need. I think it should be more than enough to validate topics during authorizePublish and authorizeSubscribe.

I prefer _not_ storing the password in the clear in the default implementation. 

If would be cool if you release an aedes-jwt integration, I think a lot of people would find it very useful.

martindederer:
@mcollina I see, that would probably work. I am usually hesitant to modify structures like the client instance that third party code provides to the consumer code since it is a potential vector for future breakage. I will give it a try though.

I agree with your objection on plain text password storage in memory. It should be avoided if possible.

I'll have to clean up the aedes-jwt stuff quite a bit before i can release it for public consumption.

drasko:
Any news on this issue? This is highly needed feature.

mcollina:
Anybody that wants to send a PR?

scagood:
:thinking: I dont think this should be part of the main library, as its acomplishable using the currently exposed auth methods. Maybe a Exensions/Plugin would be the best option to acomplish this.

In the mean time here is a really naive example:

Server using [jsonwebtoken](https://github.com/auth0/node-jsonwebtoken):
```js
const net = require('net');
const Aedes = require('aedes');
const jwt = require('jsonwebtoken');

const server = net.createServer();
const aedes = new Aedes();

const secret = process.env.SECRET_OR_KEY;

aedes.authenticate = (client, username, password, callback) => {
  if (username === 'oauth2') {
    return jwt.verify(password.toString(), secret, (error, token) => {
      if (error) {
        return callback(error, false);
      }

      client.token = token;
      return callback(null, true);
    });
  }

  return callback(null, false);
};

function checkAnyScope(client, ...requiredScopes) {
  if (typeof client.token.scope !== 'string') {
    throw new TypeError('Token contains no scope');
[...]

PR #660, opened 2021-08-18
Add username from CONNECT to client after authenticate succes

faceless2:
Re issue https://github.com/moscajs/aedes/issues/147, the basic requirement in any authentication scheme is knowing the username specified during the CONNECT phase. So I think this really should be in the core API, rather than added by users.

This patch adds a that as the `Client.username`  property.

### DIFF
--- a/docs/Client.md
+++ b/docs/Client.md
@@ -72,6 +72,15 @@ It is available only after `CONNACK (rc=0)`, otherwise it is `null` in cases:
 - after `CONNACK (rc!=0)` response
 - `connectionError` raised by aedes
 
+## client.username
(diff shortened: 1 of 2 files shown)


robertsLando:
You should add a unit test that covers this

robertsLando:
I would also know @getlarge and @mcollina thoughts on this

getlarge:
It makes sense to have the username by default in the client, personally i always add it manually :)
And yes some tests would be needed and the [types](https://github.com/moscajs/aedes/blob/main/types/client.d.ts) should be updated too.

PR #716, opened 2022-02-05
fix: enhance test on pr 583

gnought:
In PR #583, the fix is valid but the unit test does not. The test is still pass when we revert the line back to
``` js
packet.writeCallback = next
``` 
The PR mentioned Aedes uses through2 and suffers similar bugs. The claim is partially correct. We need to do a fix but it is due to the nodejs Transform callback behaviour, in our case it is `writeCallback` call. We could refer to nodejs doc https://nodejs.org/docs/latest-v16.x/api/stream.html#transform_transformchunk-encoding-callback
> The callback function must be called only when the current chunk is completely consumed. The first argument passed to the callback must be an Error object if an error occurred while processing the input or null otherwise. If a second argument is passed to the callback, it will be forwarded on to the transform.push() method. 

This PR is to fix the unit test on PR #583.


robertsLando (approved):

robertsLando:
> . We need to do a fix but it is due to the nodejs Transform callback behaviour, in our case it is writeCallback call. We could refer to nodejs doc

What could be a fix to this?

gnought:
The fix in #583 is fine, not to pass 2nd argument to in the Transform callback. :)

PR #720, opened 2022-02-11
fix: don't send retained messages when granted QoS is 128 and set granted QoS in `subscribe` event subs

### LINKED ISSUE #719: [bug] Retained messages are sent even when granted QoS is 128
Also aedes `'subscribe'` event should have the granted qos in subs array and not the requested qos of subscription request

robertsLando:
Fixes #719 

QoS 128 means the subscribe has failed so we shouldn't send any retained message in that case. Also the qos set in `subs` array passed as second argument in `subscribe` event should match the granted qos of each topic and not the qos sent in the subscribe request

### DIFF
--- a/lib/handlers/subscribe.js
+++ b/lib/handlers/subscribe.js
@@ -76,10 +76,11 @@ function _dedupe (subs) {
 function handleSubscribe (client, packet, restore, done) {
   packet.subscriptions = packet.subscriptions.length === 1 ? packet.subscriptions : _dedupe(packet.subscriptions)
   client.broker._parallel(
-    new SubscribeState(client, packet, restore, done),
-    doSubscribe,
-    packet.subscriptions,
-    restore ? done : completeSubscribe)
+    new SubscribeState(client, packet, restore, done), // what will be this in the functions
+    doSubscribe, // function to call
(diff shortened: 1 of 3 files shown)


gnought:
In MQTT 3.1/3.11 spec, SUBACK will post back QoS 128 if failed. It seems this PR allows users to set QoS 128 in SUBSCRIBE. Is it valid in MQTT 3.1/3.11 spec?

robertsLando:
> In MQTT 3.1/3.11 spec, SUBACK will post back QoS 128 if failed.

This is still happening, what do you mean? I didn't changed anythinf in the SUBACK response

> It seems this PR allows users to set QoS 128 in SUBSCRIBE. Is it valid in MQTT 3.1/3.11 spec?

As said above I didn't allow users to set QoS 128, what I did is to prevent sending retained messages if granted QoS is 128 and fix the returned subs array in aedes `subscribe` event. I didn't touched anything that changed things in that way

robertsLando:
@gnought @mcollina Please let me know if you think this is good to merge, I need it to fix a bug in production

mcollina (approved):
lgtm

robertsLando:
Grazie Matteo 👍🏼

PR #731, opened 2022-04-19
chore: drop bulk-write-stream

gnought:
This aims to reduce one dependency.
Hope I implement well for `writev` function


robertsLando (approved):

mcollina (approved):
lgtm

Issue #736, opened 2022-04-26
[bug] Nolocal flag for bridges is not respected


oldrich-s:
**System Information**
- Aedes: 0.46.3
- NodeJS: 16.14.2 LTS
- Windows Server
- Arch: x86_64

**Describe the bug**

If I am not mistaken (and it could easily be that I am mistaken) then the `nolocal` flag is not respected by Aedes:

https://github.com/mqtt/mqtt.org/wiki/bridge_protocol

> the subscriptions are nolocal (in the JMS sense), to reduce the chances of infinite message loops.

Basically what happens is that when I connect to Aedes broker as a broker and publish a message to the Aedes broker while also subscribing on '#', I receive the message that I have just published.

Could you please point me to the piece of Aedes code where you potentially do something like:

```
if (subscription.nl === true && targetClient.id === fromClient.id) {
  dontSendThePacket()
}
```

As much as I read the Aedes code, I could not find any `nl === true` or similar handling.

robertsLando:
Aedes doens't fully support brigde yet. Is it [this](https://github.com/moscajs/aedes/blob/main/lib/handlers/subscribe.js#L162) what you are looking for?

oldrich-s:
Thanks for your answer.

No, not really. If I understand the code you sent me correctly, then if the client makes a new subscription and it already exists (with the same topic) and is different from the old one then the code unsubscribes the old subscription and subscribes the new one.

What I need is that Aedes does something like the following logic:

* Aedes receives a `publish` packet from e.g `client1`
* Aedes finds all the respective subscriptions where to forward the packet to (e.g. `subscription1`, `subscription2`, `subscription3`)
* Aedes checks if the subscription (e.g. `subscription1`) is `nl === true`
* If `subscription1` is `nl === true` then
  * Aedes checks if `subscription1` is part of `client1` and if so then
    * Aedes does not forward the message to the `client1` because the `subscription1` was set to be `nolocal`

robertsLando:
@oldrich-s Would you like to send a PR for this?

IMO you could handle this in [client](https://github.com/moscajs/aedes/blob/main/lib/client.js) deliveQos method but you should give it a try. Just add a case in the `toForward` var that makes the checks you need

oldrich-s:
I can give it a shot but I think it will be more complicated than that. I need to know the origin clientId and the target clientId.

All the information about the origin client is being lost in the following function:

```
function emitPacket (packet, done) {
[...]

PR #737, opened 2022-04-26
feat: nolocal flag support for bridges

oldrich-s:
https://github.com/moscajs/aedes/issues/736

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -92,9 +92,9 @@ function Client (broker, conn, req) {
   this.deliver0 = function deliverQoS0 (_packet, cb) {
     const toForward = dedupe(that, _packet) &&
       that.broker.authorizeForward(that, _packet)
-    
+
     const isNL = _packet.clientId === that.id && _packet.nl
-    
+
     if (toForward && !isNL) {
       // Give nodejs some time to clear stacks, or we will see
       // "Maximum call stack size exceeded" in a very high load
@@ -119,9 +119,9 @@ function Client (broker, conn, req) {
     }
     const toForward = dedupe(that, _packet) &&
       that.broker.authorizeForward(that, _packet)
-    
+
     const isNL = _packet.clientId === that.id && _packet.nl
-    
+
     if (toForward && !isNL) {
       setImmediate(() => {
         const packet = new QoSPacket(toForward, that)
--- a/lib/client.js
+++ b/lib/client.js
@@ -90,6 +90,8 @@ function Client (broker, conn, req) {
   this._eos = eos(this.conn, this.close.bind(this))
 
   const getToForwardPacket = (_packet) => {
+    // Mqttv5 3.8.3.1: https://docs.oasis-open.org/mqtt/mqtt/v5.0/mqtt-v5.0.html#_Toc3901169
+    // prevent to forward messages sent by the same client when no-local flag is set
     if (_packet.clientId === that.id && _packet.nl) return
 
     const toForward = dedupe(that, _packet) &&
--- a/lib/client.js
+++ b/lib/client.js
@@ -89,13 +89,19 @@ function Client (broker, conn, req) {
   conn.on('end', this.close.bind(this))
   this._eos = eos(this.conn, this.close.bind(this))
 
-  this.deliver0 = function deliverQoS0 (_packet, cb) {
+  const getToForwardPacket = (_packet) => {
+    if (_packet.clientId === that.id && _packet.nl) return
+
     const toForward = dedupe(that, _packet) &&
       that.broker.authorizeForward(that, _packet)
 
-    const isNL = _packet.clientId === that.id && _packet.nl
+    return toForward
+  }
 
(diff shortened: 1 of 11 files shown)


lib/handlers/subscribe.js
```diff
@@ -142,13 +142,12 @@ function addSubs (sub, done) {
...
-      _packet = new Packet(_packet, broker)
-      _packet.retain = false
-      deliverFunc(_packet, cb)
-    }
+  const deliverFunc = func
+  func = function handlePacketSubscription (_packet, cb) {
+    _packet = new Packet(_packet, broker)
+    _packet.nl = nl
+    if (!rap) _packet.retain = false
+    deliverFunc(_packet, cb)
```
robertsLando, line 150: Could you explain this?

lib/client.js
```diff
@@ -116,7 +119,10 @@ function Client (broker, conn, req) {
     }
     const toForward = dedupe(that, _packet) &&
       that.broker.authorizeForward(that, _packet)
-    if (toForward) {
+
+    const isNL = _packet.clientId === that.id && _packet.nl
+
+    if (toForward && !isNL) {
```
robertsLando, line 125: Could you create a `shouldForward` method for this as it's the same for deliver0 and deliverQos?

robertsLando (changes requested):

robertsLando:
Also please add a unit test that covers this

robertsLando:
@gnought Opinions on this one?

lib/handlers/subscribe.js
oldrich-s: Originally, the code checked if rap (retain as published) was `false` (which is default) and if so then it wrapped the delivery function and injected / modified the retain flag to `false` to indicate that the message was not retained.

I have made a modification that always wraps the delivery function and always injects nl (nolocal) flag from the subscription into the packet. The `func` is then sent to mqemmiter that uses it whenever a packet is received.

So this way I make the `nl` flag from the subscription accessible in the `deliver0` and `deliverQoS` in the client.js.

lib/client.js
oldrich-s: Should be done now 👍

test/bridge.js
```diff
@@ -0,0 +1,48 @@
...
+  })
+})
+
+test('bridge client sends a publish message but shall not receive it back', function (t) {
+  const s = connect(setup(), { clientId: 'my-client-bridge-1', protocolVersion: 128 + 4 })
+  t.teardown(s.broker.close.bind(s.broker))
+
+  const handle = setTimeout(() => t.end(), 1000)
+
+  subscribe(t, s, 'hello', 0, function () {
```
robertsLando, line 35: Where do you set the no local flag here?

oldrich-s: It is set implicitly by mqtt-packet by setting `protocolVersion: 128 + 4`

oldrich-s:
> LGTM. Thanks for this PR @oldrich-s ! Is it ready to merge?

I have just added qos 1 test (just to be sure that persistence works as expected). But otherwise I believe it is ready to be merged.

robertsLando (changes requested):

lib/client.js
```diff
@@ -89,9 +89,18 @@ function Client (broker, conn, req) {
   conn.on('end', this.close.bind(this))
   this._eos = eos(this.conn, this.close.bind(this))
 
-  this.deliver0 = function deliverQoS0 (_packet, cb) {
+  const getToForwardPacket = (_packet) => {
+    if (_packet.clientId === that.id && _packet.nl) return
```
robertsLando, line 93: Could you add a reference to MQTTv5 Section here? 
```suggestion
    // Mqttv5 3.8.3.1: https://docs.oasis-open.org/mqtt/mqtt/v5.0/mqtt-v5.0.html#_Toc3901169
    // prevent to forward messages sent by the same client when no-local flag is set
    if (_packet.clientId === that.id && _packet.nl) return
```

test/bridge.js
```diff
@@ -0,0 +1,54 @@
...
+          t.end()
+        }
+      })
+  
+      s.inStream.write(packet)
+    })
+  })
+
+  test('bridge client sends a publish message but shall not receive it back, qos = ' + qos, function (t) {
+    const s = connect(setup(), { clientId: 'my-client-bridge-1', protocolVersion: 128 + 4 })
```
robertsLando, line 38:
```suggestion
    // protocolVersion 128 + 4 means mqtt 3.1.1 with bridgeMode enabled
    // https://github.com/mqttjs/mqtt-packet/blob/7f7c2ed8bcb4b2c582851d120a94e0b4a731f661/parser.js#L171
    const s = connect(setup(), { clientId: 'my-client-bridge-1', protocolVersion: 128 + 4 })
```

robertsLando (changes requested):

oldrich-s:
> LGTM. Good to merge?

yeps 👍

PR #741, opened 2022-05-12
fix(types): ClientId type definition

seriousme:
As discussed in https://github.com/moscajs/aedes-cached-persistence/pull/52#issuecomment-1124580516
Next step will be to import this definition in `aedes-persistence`

Kind regards,
Hans

### DIFF
--- a/types/packet.d.ts
+++ b/types/packet.d.ts
@@ -1,9 +1,10 @@
 import { AedesPacket } from 'aedes-packet'
 import { IConnackPacket, IConnectPacket, IPingreqPacket, IPublishPacket, IPubrelPacket, ISubscribePacket, ISubscription, IUnsubscribePacket } from 'mqtt-packet'
 
+export type ClientId = string
 export type SubscribePacket = ISubscribePacket & { cmd: 'subscribe' }
 export type UnsubscribePacket = IUnsubscribePacket & { cmd: 'unsubscribe' }
-export type Subscription = ISubscription & { clientId?: string }
+export type Subscription = ISubscription & { clientId?: ClientId }
 export type Subscriptions = { subscriptions: Subscription[] }
 
 export type PublishPacket = IPublishPacket & { cmd: 'publish' }


seriousme:
I just tried:
```typescript
export type Subscription = ISubscription & { clientId: ClientId }
```
(making `clientId` non-optional)
and this passes all tests as well !

If we can go this route we would not need any updates to `aedes-persistence` typing anymore ;-)

Happy to commit this.

Kind regards,
Hans

robertsLando (approved):

robertsLando:
> If we can go this route we would not need any updates to aedes-persistence typing anymore ;-)

I'm not sure it's 100% correct to make that property required, It depends if it's intended as a subscription request or a subscription that flows inside aedes, in first case the clientId could be undefined. Sorry I already merged this as I didn't see your comment (it's morning in Italy I think I need my cup of coffee 😆 ) , in types for example it is also used [here](https://github.com/moscajs/aedes/blob/main/types/client.d.ts#L20) and there is no need to add a clientId

seriousme:
Ok, then I suggest we overload Subscription and Subscriptions in  `aedes-persistence` like
```typescript
export type Subscription = ISubscription & { clientId: ClientId }
export type Subscriptions = { subscriptions: Subscription[] }
```
alternatively we add:
```typescript
export type PersistanceSubscription = ISubscription & { clientId: ClientId }
export type PersistanceSubscriptions = { subscriptions: Subscription[] }
```
here.

What do you think ?

Kind regards,
Hans

robertsLando:
I personally prefer the second solution, go for `PersistanceSubscription`

seriousme:
Seems like I need more of your Italian coffee as well :-)

The problem lies in the definition of ClientId in `aedes-persistence`. It is based off `Subscription["clientId"]` but it should have been based off `Client["id"]` , which was already available in Aedes: 
https://github.com/moscajs/aedes/blob/f3b4b164b702459c631e7fbbc394d6a34cfca335/types/client.d.ts#L7

So this PR is superfluous :-(

I tested on `aedes-cached-persistence` and changing:`Subscription["clientId"]` to `Client["id"]` in `aedes-persistence` resolves the `tsd` error you got. It still complains about the missing `@types/node`

My suggestion is the following:
1) I submit a PR to revert this PR on `aedes`
2) I submit a PR to change:`Subscription["clientId"]` to `Client["id"]` and add `tsd` testing on `aedes-persistence` 
3) I submit a PR to update `aedes-persistence` to the latest version, add  `@types/node` as a Dev dependency and add `tsd` testing on `aedes-cached-persistence` 

Ok ?

Kind regards,
Hans

PR #768, opened 2022-07-31
feat: add a prefix in aedes module and drop ts functions

gnought:
A breaking change in TS
- drop Server() and Aedes() ts export functions
- add a prefix in aedes modules 

Sample usage
``` js
import Server, { Aedes } from 'aedes:server'
// the above is same as below
// import Server, { Aedes } from 'aedes'

let broker = new Server()
broker = new Aedes()
```

### DIFF
--- a/aedes.d.ts
+++ b/aedes.d.ts
@@ -1,9 +1,3 @@
-import { Aedes, AedesOptions } from './types/instance'
-
-export declare function aedes (options?: AedesOptions): Aedes
-export declare function Server (options?: AedesOptions): Aedes
-
-export * from './types/instance'
-export * from './types/packet'
-export * from './types/client'
-export default aedes
+/// <reference path="./types/instance.d.ts" />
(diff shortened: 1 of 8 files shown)


robertsLando (approved):
LGTM

robertsLando:
Should we go for 1.0.0 ?

gnought:
The core doesn't change. IMO it is sufficient to bump a minor version.

A LWT authorise fix is coming, mind to wait?

robertsLando:
Sure 👍🏼 Glad to see you back :)

robertsLando:
This seems to have break something on my instance:

```ts
import { Client } from 'aedes:client';
```

Error: 

```bash
error TS2614: Module '"aedes"' has no exported member 'Client'. Did you mean to use 'import Client from "aedes"' instead?

14 import { Client } from 'aedes';
```

cc @gnought

Issue #767, opened 2022-07-30
[bug] Failure to authorize old wills


friesendrywall:
**System Information**
- Aedes: 0.47.0
- NodeJS: Any
- OS: Any

**Describe the bug**
Wills on startup are not authorized

**To Reproduce**
Add test to will.js and npm run test

**Expected behavior**
Wills should be authorized

This seems to me that it will require adding to the persistence layer, as extra client information will need to be added, namely any important information like username, etc that gets passed to the authorizePublish handler.

Here is a failing test example.

```
test('authorize old will in case of a crash', function (t) {
  t.plan(3)

  const persistence = memory()
  const will = {
    topic: 'mywill',
    payload: Buffer.from('last will'),
    qos: 0,
    retain: false
  }

  persistence.broker = {
    id: 'anotherBroker'
  }

  persistence.putWill({
    id: 'myClientId42'
  }, will, function (err) {
    t.error(err, 'no error')

    const interval = 10 // ms, so that the will check happens fast!
    const broker = aedes({
      persistence: persistence,
      heartbeatInterval: interval
    })
    let authorized = false
    broker.authorizePublish = function (client, packet, callback) {
      authorized = true
      callback(null)
    }

    t.teardown(broker.close.bind(broker))

    const start = Date.now()

    broker.mq.on('mywill', check)

    function check (packet, cb) {
      broker.mq.removeListener('mywill', check)
      t.ok(Date.now() - start >= 3 * interval, 'the will needs to be emitted after 3 heartbeats')
      t.equal(authorized, true, 'topic authorized')
      broker.mq.on('mywill', function (packet) {
        t.fail('the will must be delivered only once')
      })
      cb()
    }
  })
})
```

friesendrywall:
Why was this closed?  I'm a bit confused, because the present code doesn't authorize these re-sent wills, or maybe that is ok.

gnought:
The #769 is merged to fix this issue. @friesendrywall

PR #769, opened 2022-08-01
fix: authorize old wills

gnought:
This is a response to #767, to authorise old wills be published. The `client` will be `null` in `authorizePublish` in this case.

### DIFF
--- a/aedes.js
+++ b/aedes.js
@@ -115,27 +115,25 @@ function Aedes (opts) {
   }
 
   function checkAndPublish (will, done) {
-    const needsPublishing =
-      !that.brokers[will.brokerId] ||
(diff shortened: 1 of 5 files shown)


robertsLando (approved):
LGTM. I think it makes sense to add to docs a notice about broker id, by default it has a random id and this could break the will restore logic

gnought:
@robertsLando would be mind to elaborate about the will restore logic you are referring to?

robertsLando:
@gnought The one we are speaking about here: https://github.com/moscajs/aedes/pull/756

gnought:
> @gnought The one we are speaking about here: #756

I see. This fix does not relate to the will restore logic. It adds an authorisation logic before aedes publishes the old wills.

robertsLando:
Yep not related to this was just to mention that :)

Issue #187, opened 2018-02-26
Improve and add more examples


Domvel:
The documentation is not clearly how to create a instance of aeded and subscribe a channel.
Please improve the example with a simple subscripte. 

Client side: 
```js
var mqtt = require('mqtt');
var client = mqtt.connect('mqtt://my-broker-ip');

client.on('connect', function () {
  client.subscribe('test');
  client.publish('test', 'Hello mqtt');
})

client.on('message', function (topic, message) {
  // message is Buffer
  console.log(message.toString());
  client.end();
});
```

How to setup the server side to get the message?

PS: I like this software. A lightweight fast installed mqtt broker for node js. 👍

Domvel:
Here's are suggestion of a new extended example: 

```js
var aedes = require('aedes')()
var server = require('net').createServer(aedes.handle)
var port = 1883

server.listen(port, function() {
  console.log('server listening on port', port);
});

aedes.subscribe('test', function(packet, cb) {
  console.log('Published', packet.payload.toString());
});
```

Domvel:
Another (dirty) example of conversation:

`Server`
```js
var aedes = require('aedes')()
var server = require('net').createServer(aedes.handle)
var port = 1883

server.listen(port, function() {
  console.log('server listening on port', port);
});

aedes.subscribe('fromClient', function(packet, cb) {
  console.log('Published', packet.payload.toString());

  aedes.publish({
    cmd: 'publish',
    qos: 2,
    topic: 'fromServer',
    payload: new Buffer('Thank you!'),
    retain: false
  });
});
```

`Client`
```js
let mqtt = require('mqtt');
let client = mqtt.connect('mqtt://server.ip');

client.on('connect', function () {
  client.subscribe('fromServer');

  client.publish('fromClient', new Date().toISOString());

  setInterval(() => {
    client.publish('fromClient', new Date().toISOString());
  }, 1000);
})

client.on('message', function (topic, message) {
  // message is Buffer
  console.log('Received a message', topic, message.toString());
  //client.end();
});
```

Domvel:
A simple login / authentification: 

`server`
```ts
aedes.authenticate = function(client, username, password, callback) {
  console.log('Client connected. Login ...');
  callback(null, username === 'domvel');
}
```

`client`
```js
let client = mqtt.connect('mqtt://192.168.1.164', {
  username: 'domvel',
  password: 'secret'
});
```

Not handled: Errors on `not authenticated`.

[...]

PR #793, opened 2022-12-04
Add some basic examples

insanity54:
I think basic examples are necessary to grasp usage. When learning aedes, I didn't see any non-cluster examples in the examples dir so I had to dig through the issue tracker to find some.

This PR addresses https://github.com/moscajs/aedes/issues/187. There are some good examples in that issue thread, which I've copied and placed in the examples directory.

### DIFF
--- a/examples/basic-client.js
+++ b/examples/basic-client.js
@@ -1,0 +1,31 @@
+// example taken from https://github.com/moscajs/aedes/issues/187#issuecomment-511886920
+
+// Client
+const SERVER = 'mqtt://localhost'
+const TOPIC = 'main'
+
+let mqtt = require('mqtt')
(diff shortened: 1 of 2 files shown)


robertsLando:
Isn't this enough: https://github.com/moscajs/aedes/blob/main/docs/Examples.md ?

insanity54:
I don't think so. It's missing an example client.

robertsLando:
The client example depends on the client library you use, for example mqttjs provides it's own example in [docs](https://github.com/mqttjs/MQTT.js#example)

insanity54:
Okay, I understand. Maybe the examples in this PR is out of scope of aedes server.

I still think aedes has a documentation issue; I would like to improve them. I'm going to open a separate issue.

Issue #801, opened 2022-12-27
[bug] TypeScript error due to incompatible aedes.handle function for Node net module v16


haci-xplora:
**System Information**
- Aedes: 0.48.1
- NodeJS: 16.18.1
- TypeScript: 4.9.4
- OS: MacOS
- Arch: Apple M1

**Describe the bug**
Function `aedes.handle` is not compatible with listener for `net.createServer`. So, it causes TypeScript compile error.

Sample code:

```typescript
import Aedes from 'aedes';
import { createServer } from 'net';
const mqttServer = new Aedes();
const netServer = createServer(mqttServer.handle);
netServer.listen(1883, () => {
  console.info('ready at 1883');
});
```

Error:

```plain
src/index.ts:4:32 - error TS2769: No overload matches this call.
  Overload 1 of 2, '(connectionListener?: ((socket: Socket) => void) | undefined): Server', gave the following error.
    Argument of type '(stream: Connection, request: IncomingMessage) => Client' is not assignable to parameter of type '(socket: Socket) => void'.
  Overload 2 of 2, '(options?: ServerOpts | undefined, connectionListener?: ((socket: Socket) => void) | undefined): Server', gave the following error.
    Type '(stream: Connection, request: IncomingMessage) => Client' has no properties in common with type 'ServerOpts'.

4 const netServer = createServer(mqttServer.handle);
                                 ~~~~~~~~~~~~~~~~~
Found 1 error.
```

**To Reproduce**
Steps to reproduce the behavior is not much different than the sample code on [Examples](https://github.com/moscajs/aedes/blob/main/docs/Examples.md) page

1. try to compile project with sample code

**Expected behavior**
Project should compile as expected without any errors.

robertsLando:
@haci-xplora this is a type issue, could you please send a PR to fix it?

haci-xplora:
hi Robert.

the type of `handle` function is based on the signature of implementation, so it is correct. but if request parameter (`IncomingMessage`) is not in use (as I can see it is not used), it should be omitted. that is why it should not be assigned as listener of net server. shall I do that and remove request parameter?

or we can make it optional/nullable.

otherwise, multiple adapter clients are needed for `net` server, and other servers like `websocket` server.
see [Client.md](https://github.com/haci-xplora/aedes/blob/main/docs/Client.md#clientreq)
it says request parameter is for that.

one class cannot handle those variations well.

let me create a PR and discuss further.

haci-xplora:
[...]

PR #802, opened 2022-12-28
fix: typescript issue related to `aedes.handle` and client

haci-xplora:
ref: [issue 801](https://github.com/moscajs/aedes/issues/801)

### DIFF
--- a/aedes.js
+++ b/aedes.js
@@ -52,7 +52,7 @@ function Aedes (opts) {
     concurrency: opts.concurrency,
     matchEmptyLevels: true // [MQTT-4.7.1-3]
   })
-  this.handle = function handle (conn, req = null) {
+  this.handle = function handle (conn, req) {
     conn.setMaxListeners(opts.concurrency * 2)
     // create a new Client instance for a new connection
     // return, just to please standard
--- a/aedes.js
+++ b/aedes.js
@@ -52,7 +52,7 @@ function Aedes (opts) {
     concurrency: opts.concurrency,
     matchEmptyLevels: true // [MQTT-4.7.1-3]
   })
-  this.handle = function handle (conn, req) {
+  this.handle = function handle (conn, req = null) {
     conn.setMaxListeners(opts.concurrency * 2)
     // create a new Client instance for a new connection
     // return, just to please standard
--- a/lib/client.js
+++ b/lib/client.js
@@ -15,7 +15,7 @@ const { through } = require('./utils')
 
 module.exports = Client
 
-function Client (broker, conn, req = null) {
+function Client (broker, conn, req) {
   const that = this
 
   // metadata
--- a/lib/client.js
+++ b/lib/client.js
@@ -15,7 +15,7 @@ const { through } = require('./utils')
 
 module.exports = Client
 
-function Client (broker, conn, req) {
+function Client (broker, conn, req = null) {
   const that = this
 
   // metadata
--- a/types/client.d.ts
+++ b/types/client.d.ts
@@ -9,7 +9,7 @@ declare module 'client' {
     clean: Readonly<boolean>
     version: Readonly<number>
     conn: Connection
-    req?: IncomingMessage | null
+    req?: IncomingMessage
     connecting: Readonly<boolean>
     connected: Readonly<boolean>
     closed: Readonly<boolean>
--- a/types/client.d.ts
+++ b/types/client.d.ts
@@ -9,7 +9,7 @@ declare module 'client' {
     clean: Readonly<boolean>
     version: Readonly<number>
     conn: Connection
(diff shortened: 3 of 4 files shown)


robertsLando (changes requested):
No need to set it to `null` simply add the `?` parameter, when omitted it will be undefined

haci-xplora:
to specify optional parameter in JavaScript, I have to assign it to `null` and then go and fix that in TypeScript type definition accordingly.

what do you think?

haci-xplora:
also we should not use expressions like `{} as IncomingMessage` in some tests. we can use `null` now.

shall I fix that as well?

types/instance.d.ts
```diff
@@ -66,7 +66,7 @@ declare module 'aedes' {
     brokers: Readonly<Brokers>
 
     constructor(option?: AedesOptions)
-    handle: (stream: Connection, request: IncomingMessage) => Client
+    handle: (stream: Connection, request?: IncomingMessage | null) => Client
```
robertsLando, line 69:
```suggestion
    handle: (stream: Connection, request?: IncomingMessage) => Client
```

types/client.d.ts
```diff
@@ -9,7 +9,7 @@ declare module 'client' {
     clean: Readonly<boolean>
     version: Readonly<number>
     conn: Connection
-    req?: IncomingMessage
+    req?: IncomingMessage | null
```
robertsLando, line 12:
```suggestion
    req?: IncomingMessage
```

lib/client.js
```diff
@@ -15,7 +15,7 @@ const { through } = require('./utils')
 
 module.exports = Client
 
-function Client (broker, conn, req) {
+function Client (broker, conn, req = null) {
```
robertsLando, line 18:
```suggestion
function Client (broker, conn, req) {
```

aedes.js
```diff
@@ -52,7 +52,7 @@ function Aedes (opts) {
     concurrency: opts.concurrency,
     matchEmptyLevels: true // [MQTT-4.7.1-3]
   })
-  this.handle = function handle (conn, req) {
+  this.handle = function handle (conn, req = null) {
```
robertsLando, line 55:
```suggestion
  this.handle = function handle (conn, req) {
```

robertsLando (changes requested):
I dunno what you mean, does using the changes I suggest here creates errors somewhere?

aedes.js
haci-xplora: as mentioned in the docs, this is a fix for the optional parameter which is only used for websockets.

lib/client.js
haci-xplora: as mentioned in the docs, this is a fix for the optional parameter which is only used for websockets.

types/client.d.ts
haci-xplora: this change is required based on the changes required in the javascript code.

types/instance.d.ts
haci-xplora: this change is required based on the changes required in the javascript code.

haci-xplora:
if you do not want to change javascript files, but only typescript files, that would work. but then the discrepancy between the code and the documentation stays.

lib/client.js
robertsLando: @haci-xplora I know that but there is still no reason to set it to null as it will be undefined when omitted

aedes.js
robertsLando: @haci-xplora I know that but there is still no reason to set it to null as it will be undefined when omitted

types/client.d.ts
robertsLando: See previous answer

types/instance.d.ts
robertsLando: See previous answer

aedes.js
haci-xplora: without using null (a default value), by looking at the code, a dev can falsely assume that it is a required parameter.

haci-xplora:
changes on js files undone.

aedes.js
robertsLando: req is not used in code anywhere, it is just set so outside aedes users can access it, when the type is ok in typescript vscode will tell them it is optional and that may not exists so the problem doesn't exist

haci-xplora:
almost undefined vs null

it's been an interesting discussion

robertsLando (approved):

robertsLando:
@haci-xplora I think it may make sense to also add a types test [here](https://github.com/moscajs/aedes/blob/main/test/types/aedes.test-d.ts#L128) that also calls handle without the req to prevent regressions

robertsLando:
Supersided by #812

PR #806, opened 2023-01-12
fix(types): update username and password type definition

adipascu:
The username and password variables have the undefined type when not set in the MQTTX client.


robertsLando (approved):

robertsLando (changes requested):
Please fix types tests issues

robertsLando (approved):

Issue #878, opened 2023-07-21
Difficult to use in ES6 module package.


peeveen:
In my TypeScript project, this code works ok ...

```
import Aedes from "aedes"
const aedes: Aedes = new Aedes();
```

... until I add `"type": "module"` to my `package.json`.
It now shows the errors `Cannot use namespace 'Aedes' as a type` and `the expression is not constructable`.

This is only workaround I could figure out:

```
const aedes: Aedes.default = Aedes.createBroker()
```

robertsLando:
Please consider opening a PR to fix this issue

hjdhjd:
As I mentioned [here](https://github.com/moscajs/aedes/pull/879#issuecomment-1824622434), and may be useful for others...this PR breaks TypeScript in straightforward use cases with simple tsconfigs.

I would suggest this PR is reverted or updated to correct the errors. Removing the code introduced in the PR fixes the issue.

robertsLando:
@hjdhjd Would you like to submit a PR to fix that?

hjdhjd:
Happy to, but it would be to revert the changes made in the prior PR. If you're good with that, I will be happy to submit it.

robertsLando:
isn't there a way to keep the changes and fix what's not working?

hjdhjd:
The changes **are** the problem - they shouldn't be there in the first place because the prior code was correct. Here's a helpful stackoverflow answer: https://stackoverflow.com/questions/55506545/export-assignment-and-exporting-types-in-typescript-ambient-module

In experimenting with aedes for a couple of days it seems many of the TypeScript typings are either incorrect or ambiguously defined. I'll take a closer look all around and put together a PR to correct some of them.

robertsLando:
@hjdhjd I agree types can be imrpved a lot, go for the PR I will take a look :+1

hjdhjd:
> @hjdhjd I agree types can be imrpved a lot, go for the PR I will take a look 👍🏼

Submitted.

getlarge:
I just encountered a similar issue when trying to import in Typescript a module exported with `module.exports = TheModule`.
Since `TheModule` (in that case, `ClinicBubbleprof` from `@clinic/bubbleprof`) has no type definitions, I created them and the correct way to export it was by using `export = TheModule`.
Then in the TS file, the import was done with `import TheModule = require('path-to-the-module');`
To keep on using the `import X from x` syntax, i had to set `esModuleInterop=true` in `compilerOptions`.

[...]

PR #879, opened 2023-07-23
added module declaration

peeveen:
Suggested fix for https://github.com/moscajs/aedes/issues/878


robertsLando (commented):
Could you add a typescript test for this? Just to check it's working as expected

peeveen:
> Could you add a typescript test for this? Just to check it's working as expected

I spent a few hours trying this. I tried to create a whole other test "project" inside the `test` folder, with it's own `package.json` with `"type": "module"` in it, and make a relative local reference to the parent `aedes` package as a dependency, which became a circular folder symlink/shortcut ... but I think node/tap/npm/tsc/something is getting confused, because I can't get my test code to stop complaining about "cannot use namespace as type" unless I add `"type": "module"` to the __main project__ `package.json` ... which causes all manner of other problems.

Basically, I gave up.

But adding those three lines into the `node_modules/aedes/aedes.d.ts` file in my project is making everything work beautifully.

If the lack of an internal test is a big problem, just reject this. I'll keep using the workaround.

robertsLando (approved):
It's ok, no worries

hjdhjd:
This PR breaks TypeScript modules with the following error:

```
node_modules/aedes/aedes.d.ts:11:3 - error TS2666: Exports and export assignments are not permitted in module augmentations.
```

In general this isn't a good idea, as I understand it. As of right now, Aedes is unusable in even the simplest of TypeScript projects.

PR #918, opened 2023-11-26
Fix module exports for TypeScript for ES6 compliance.

hjdhjd:
Addresses https://github.com/moscajs/aedes/issues/878#issuecomment-1825783310 and corrects the export declaration in TypeScript.

### DIFF
--- a/aedes.d.ts
+++ b/aedes.d.ts
@@ -1,12 +1,5 @@
-import Aedes, { AedesOptions } from './types/instance'
+export * from './types/instance';
+export * from './types/packet';
+export * from './types/client';
 
-export declare function createBroker (options?: AedesOptions): Aedes
-
-export * from './types/instance'
-export * from './types/packet'
-export * from './types/client'
-export default Aedes
-
-declare module 'aedes' {
-  export = Aedes
-}
+export { default } from './types/instance';


aedes.d.ts
```diff
@@ -1,12 +1,5 @@
-import Aedes, { AedesOptions } from './types/instance'
+export * from './types/instance';
+export * from './types/packet';
+export * from './types/client';
 
-export declare function createBroker (options?: AedesOptions): Aedes
```
robertsLando, line 3: why did you removed this?

aedes.d.ts
```diff
@@ -1,12 +1,5 @@
...
-
-export * from './types/instance'
-export * from './types/packet'
-export * from './types/client'
-export default Aedes
-
-declare module 'aedes' {
-  export = Aedes
-}
+export { default } from './types/instance';
```
robertsLando, line 5: I'm not sure this is correct

robertsLando (changes requested):

hjdhjd:
It is correct, and I've confirmed its use.

To explain it more verbosely:

```
// Since Aedes is a default export from './types/instance', we can do this:
export { default } from './types/instance';

// If Aedes was NOT the default export, we would need to create an alias for the default export like so:
// export { default as Aedes } from './types/instance';
```

Agree on `createBroker`...I can add it back in, though I suspect it's superfluous. Let me know. The bug this fixes prevents me from using the latest Aedes and I'm forced to use 0.49 in the meantime.

robertsLando:
Could you add a typescript test that covers this to prevent any possible breaking change?

hjdhjd:
> Could you add a typescript test that covers this to prevent any possible breaking change?

Added an additional check for createBroker, but there's not much more I can do here. Do let me know if you're going to accept this or not - I'd like to get this PR closed out either way. Thanks.

robertsLando (approved):
Please fix lint issues.

robertsLando (changes requested):
Lint issues, fix them: `npm run lint:fix`

hjdhjd:
Linting done.

robertsLando (approved):

PR #930, opened 2024-01-12
feat: add `opts.keepaliveLimit`

Gianluca-Casagrande-Stiga:
Hi,

this is a PR for this discussion: https://github.com/moscajs/aedes/discussions/865

### DIFF
--- a/lib/handlers/connect.js
+++ b/lib/handlers/connect.js
@@ -36,7 +36,8 @@ const errorMessages = [
   'identifier rejected',
   'Server unavailable',
   'bad user name or password',
-  'not authorized'
+  'not authorized',
+  'keep alive limit exceeded'
 ]
 
 function handleConnect (client, packet, done) {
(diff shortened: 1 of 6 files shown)


lib/handlers/connect.js
```diff
@@ -66,9 +67,14 @@ function init (client, packet, done) {
   if (packet.protocolVersion === 3 && clientId.length > client.broker.maxClientsIdLength) {
     returnCode = 2
   }
+  // check if the client keepalive is compatible with broker settings
+  if (client.broker.keepaliveLimit && (!packet.keepalive || packet.keepalive > client.broker.keepaliveLimit)) {
```
robertsLando, line 71: Shouldn't be packet keepalive `0` allowed?

robertsLando (changes requested):

lib/handlers/connect.js
Gianluca-Casagrande-Stiga: client.keepalive 0 means "infinite keepalive", so if I set a broker.keepaliveLimit, I would like to block the clients with keepalive 0

Gianluca-Casagrande-Stiga: by default, broker.keepaliveLimit is set to 0: this makes no changes in aedes behaviour, because no client.keepalive check is done

robertsLando (approved):

PR #957, opened 2024-06-07
fix: delete will from persistence on disconnect

bmeverett:
## Problem 
I was using aedes-persistence-redis and noticed that if clients connect with a will message then the keys weren't being deleted on a clean disconnect

## Solution
The keys should always be removed from the persistence regardless if the the client disconnects cleanly or not.  When the client disconnects, the will no longer should matter so we can delete it. If the client connects again with another will, then we can retain that one.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -313,7 +313,14 @@ Client.prototype.close = function (done) {
           }, noop)
         }
       })
+    } else if (will) {
+      // delete the persisted will even on clean disconnect
+      that.broker.persistence.delWill({
+        id: that.id,
+        brokerId: that.broker.id
+      }, noop)
     }
+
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -313,7 +313,14 @@ Client.prototype.close = function (done) {
           }, noop)
         }
       })
+    } else if (will) {
+      // delete the persisted will even on clean disconnect
```
robertsLando, line 317:
```suggestion
      // delete the persisted will even on clean disconnect https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc385349232
```

Issue #955, opened 2024-05-27
[bug] Not all messages are consumed when dedupe is enabled


zagadheesh:
**System Information**
- Aedes: 0.51.1
- NodeJS: 16.15.0
- OS: Mac OS 14.5
- Arch: x86_64

**Describe the bug**
I'm currently experiencing an issue while running Aedes alongside an MQTT Java client using example.js (attached) with no modifications to the aedes source code. Despite running both the consumer and producer, it seems that certain messages aren't being consumed. Interestingly, after commenting out the 'dedupe' call in the client.js file of the Aedes source code, the consumer is able to successfully consume all messages.

**To Reproduce**
1. Run the broker using example.js on aedes 0.51.1 source code
2. Run AedesSampleConsumer.java
3. Run AedesSampleProducer.java

**Expected behavior**
Consumer should consume all the messages produced.

[aedes-problem-demo.zip](https://github.com/moscajs/aedes/files/15452958/aedes-problem-demo.zip)

robertsLando:
@zagadheesh I have actually no time to dig into this, could you do try to draft a PR to fix the issue? I already fixed dedupe a while ago for the same reason and the problem was the persistence used

zagadheesh:
@robertsLando The proposed pull request would involve commenting out the 'dedupe' call in client.js. After conducting tests on release 0.42.6 without altering any source code, I observed that the consumer successfully consumed all messages. However, beyond this change, I haven't delved into the entirety of the source code. Any suggestions would be greatly appreciated!

robertsLando:
Commenting the dedupe fn is not the solution, it is there for a reason and it should stay there, we need to find out why the messages are not deduped correctly

zagadheesh:
Reproduced the same in node.js. Please find attched zip file which contains example.js which starts the broker with required configuration along with producer.js and consumer.js files.

[sample.zip](https://github.com/moscajs/aedes/files/15455110/sample.zip)

mvandevalk:
The issue with this bug seems to revolve around the order of the returned value of the brokerCounter. It appears reproducible when using MongoDB as the persistence type. My assumption is that we do not want to ignore out-of-order counts but rather counts that have already been processed.

For example:
- First packet: brokerId = 1, brokerCounter = 10 -> return true
- Second packet: brokerId = 1, brokerCounter = 9 -> return true
- Third packet: brokerId = 1, brokerCounter = 11 -> return true
[...]

PR #972, opened 2024-08-15
fix: #955 - Updated dedupe to check history of sent packets

mvandevalk:
Updated the dedupe function to checks if a packet has already been seen before, based on the packet's broker ID and counter. If the packet is a duplicate (already seen before), it is not forwarded again. The temporary list for dedupe is configurable as part of the init options, defaults to 100.

This bug occurs with a mongoDB persistence configuration and have not been able to reproduce with a default persistence settings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ node_modules
 .lock-wscript
 .vscode
 *code-workspace
+.idea
 
 # 0x
 flamegraph.html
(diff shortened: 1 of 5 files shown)


mcollina (changes requested):
Thanks for opening a PR! Can you please add a unit test?

robertsLando:
@mvandevalk I think this should be fixed on mongo persistence instead, if this happens it means the packets are not processed in the correct order

Issue #928, opened 2024-01-08
[bug] duplicated closed clients


Gianluca-Casagrande-Stiga:
**System Information**
- Aedes: 0.50.1
- Node: 18.19.0
- OS/Arch: (Docker) node:18-alpine
- Configuration: Aedes cluster with Redis mq-emitter (6 Aedes pods + 1 Redis pod)

**Describe the bug**
The Aedes.clients object can have many closed clients that are not removed from the list after disconnection. The number of closed clients increase with usage, potentially causing memory leak.
It's only a broker side issue: on the client side connection is ok.
The closed client list can be cleared by broker restart, or manually deleting the elements of broker.clients where closed = true.

Unfortunately I haven't been able to reproduce the problem yet on dev env.
Studying my production logs, I've found that the closed client creation in the list is related to the error "connection closed" on write.js.

I think the issue is related to many disconnection/reconnection of the same client in a cluster env. These connections could be managed by different brokers at the same time. The result is, for example, that the same client ID is closed in broker1.clients and it is connected in broker2.clients. 

**Expected behavior**
A closed client shouldn't remain in the broker client list.

**Workaround**
When the closed client list is too long, I execute a function to clear the list.

**Question**
Does anyone have the same problem?
I'll keep working on this and I will write here if there will be any updates.

robertsLando:
@Gianluca-Casagrande-Stiga The only place where client is removed is:

https://github.com/moscajs/aedes/blob/981d071e189309797c3227c9f105862cc23c2576/lib/client.js#L326-L328

That calls:

https://github.com/moscajs/aedes/blob/981d071e189309797c3227c9f105862cc23c2576/aedes.js#L310-L318

I'm not sure if in the first if the `that._authorized` check could cause some memory leaks

PR #979, opened 2024-08-25
fix: memory leak in connected clients counter

uladzimir-danko:
This pr is related to the https://github.com/moscajs/aedes/issues/928 
The made fix **didn't decrease connectedClients** counter, which lead to incorrect amount of clients provided by broker.

One more issue is related to constant increase of subscriptions. My broker has a lot of unique clients per day with short keepAliveTimeout and frequent reconnection and subscription rate. I suspect that the problem is in the **authorizeSubscribe** method which can be async. This may lead to subscription for closed clients. In this case broker will never unsub them. 

Unfortunately, I am limited in experiments on production to prove my idea, but the proposed fix seems reasonable and harmless to me.

### DIFF
--- a/aedes.js
+++ b/aedes.js
@@ -172,7 +172,7 @@ function Aedes (opts) {
     if (that.clients[clientId] && serverId !== that.id) {
       if (that.clients[clientId].closed) {
         // remove the client from the list if it is already closed
-        delete that.clients[clientId]
+        that.deleteClient(clientId)
         done()
       } else {
         that.clients[clientId].close(done)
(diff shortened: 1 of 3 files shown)


aedes.js
```diff
@@ -172,7 +172,7 @@ function Aedes (opts) {
     if (that.clients[clientId] && serverId !== that.id) {
       if (that.clients[clientId].closed) {
         // remove the client from the list if it is already closed
-        delete that.clients[clientId]
+        that.deleteClient(clientId)
```
robertsLando, line 175: I think the leak was here, we were not decresing clients counter

robertsLando (approved):

PR #1027, opened 2025-05-05
chore: add type commonjs

gnought:
Followed the guideline in the [ESM](https://nodejs.org/api/esm.html#esm_package_json_type_field), we explicitly state module type.


robertsLando (approved):
Glad to see you @gnought :) LGTM

PR #1032, opened 2025-06-19
chore: migrate to ES6 classes

seriousme:
This PR migrates Aedes to ES6 class syntax.
This should help in further refactoring.
I tried to migrate everything that looked like a class but I might have missed one as `this` is also used in functions that are not constructors but whose this will be set by `parallel` etc.

Kind regards,
Hans

### DIFF
--- a/aedes.js
+++ b/aedes.js
@@ -1,7 +1,6 @@
 'use strict'
 
 const EventEmitter = require('events')
-const util = require('util')
 const parallel = require('fastparallel')
 const series = require('fastseries')
 const { v4: uuidv4 } = require('uuid')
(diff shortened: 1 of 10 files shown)


seriousme:
@robertsLando 
the constructor of Aedes immediatly starts running,e.g. it contains:

```js
  this._heartbeatInterval = setInterval(heartbeat, opts.heartbeatInterval)
```
 and:

```js
  that.publish({
      topic: birthTopic,
      payload: bufId
    }, noop)
```

My suggestion would be to migrate these actions to a separate async method as discussed before where we also can initialize async persistence as those action should imo come after persistence has been initialized.

We already have Aedes.createBroker() which we could repurpose for that.

As long as we are using the callback persistence we can automagically call this in the createBroker factory.
In the next major version the way to start Aedes could then be:

```js
const { Aedes } = require("aedes")
const aedes = await Aedes.createBroker(opts)
```
What do you think?

robertsLando:
@seriousme can I merge?

seriousme:
Yes you can.

PR #1035, opened 2025-07-04
feat: automate benchmarking

seriousme:
This PR automates the running of sender/receiver benchmarks to make it easier to compare between versions.

KInd regards,
Hans


robertsLando (changes requested):
What about adding a workflow that runs benchmarks on every PR push and then comments on PR with results? 

I would not Push examples folder BTW

seriousme:
> What about adding a workflow that runs benchmarks on every PR push and then comments on PR with results?
> 
> I would not Push examples folder BTW

Just removed the examples folder.
Running benchmark on every push is possible but it depends on how stable the performance of github runners is. And so far I have seen quite some variation. So I'm  a bit reluctant to put this in as it might give a false sense of exactness.

seriousme:
> > 1. make QoS a parameter to reduce the number of scripts
> 
> Agree 👍🏼

Sofar I kept the changes to the existing scripts as small as possible to avoid any errors there. But this is easy to fix.

> 
> > 2. publish HR-time (like pingpong.js does) and let the receiver report on latency as well
> 
> 👍🏼
>

Ok, easy to add. This would also remove the need for pingpong.js

> > 3. encode a sequence counter and check for double/missing packets.
> 
> 👍🏼

Ok, easy to add.

> 
> > 4. add deeper analysis  on statistical variations
> 
> We could keep this for another PR maybe but not so important

From what I have seen (and I'm no expert on the subject) statistics is a true art and things like confidence levels and overfitting might produce results that look great but are actually skewed. Hence my caution with the subject.

> 
> > 5. utilize techniques like described in: https://github.com/nodejs/node/blob/main/doc/contributing/writing-and-running-benchmarks.md
> 
> What do you have in mind specifically about this?

I designed the CSV format to align with [node-benchmark-compare](https://www.npmjs.com/package/node-benchmark-compare) (although its only documented by examples 😞 ) so that we might be able to leverage these tools as well if we would want to do so in the future.

> 
> > 6. adding a [diagnostic channel](https://nodejs.org/api/diagnostics_channel.html#diagnostics-channel) to Aedes so people can get runtime statistics
> 
> There is aedes-stats package that does this?

Not exactly:  [Aedes-stats](https://www.npmjs.com/package/aedes-stats)  attaches hooks that publish stats to MQTT.
And then you need a client that is able to make sense out of this.

The idea behind diagnostic channel (from what I read about it 😉 ) is that its more lightweight ("zero overhead") and more standardized so that tools like Inspector and others can tap into the data as well. See also: https://medium.com/@farhad.gulizada/diagnostics-channel-in-node-js-intermediate-series-78741cb8ab8e

Kind regards,
Hans

robertsLando:
> The idea behind diagnostic channel (from what I read about it 😉 ) is that its more lightweight ("zero overhead") and more standardized so that tools like Inspector and others can tap into the data as well. See also: [medium.com/@farhad.gulizada/diagnostics-channel-in-node-js-intermediate-series-78741cb8ab8e](https://medium.com/@farhad.gulizada/diagnostics-channel-in-node-js-intermediate-series-78741cb8ab8e)

I sincerely never heard about that since now, I'm ok for that but I would keep it for another PR? Could you create an issue to keep track of that? :)

seriousme:
I added:

> 1) make QoS a parameter to reduce the number of scripts

Done
> 2) publish HR-time (like pingpong.js does) and let the receiver report on latency as well

Not done, as calculating the stats slows down reading of the queue, meaning that latencies increase more and more over time . So I kept pingpong, but improved it a bit.

> 3) encode a sequence counter and check for double/missing packets.

Done

> 4) add deeper analysis on statistical variations

Done on pingpong.js.  By default it will calculate min/max/mean/median/95th percentile/98th percentile. All in such a way that it has relatively limited performance impact while sacraficing a small bit of accuracy (rounding to ms) 
`runBenchmark.js` currently reports on the 95th percentile as this is typically more relevant than mean while ignoring incidental spikes

> 5) utilize techniques like described in: https://github.com/nodejs/node/blob/main/doc/contributing/writing-and-running-benchmarks.md

No change on this.

> 6) adding a [diagnostic channel](https://nodejs.org/api/diagnostics_channel.html#diagnostics-channel) to Aedes so people can get runtime statistics

No change on this.

Kind regards,
Hans

seriousme:
Once this PR is in we could try to create a github action that:
1) checks out the main branch
2) runs `npm install`
3) runs `node runBenchmark.js > result.main.csv`    
4) checks out the current branch of the PR, while keeping result.main.csv 
5) runs `npm install`
6) runs `node runBenchmark.js > result.branch.csv `   
7) runs `cat result.main.csv result.branch.csv | node report.js > result.md` 
8) posts the contents of result.md to the PR that started the action

This way we at least know that the benchmark of main and current branch run on the same virtual machine setup, which should make them more comparable.
Downside of this approach is that it takes some more time before the action completes.
The alternative is to save the `result.main.csv` in the source tree and then only calculate `result.main.csv` but that be running on a more/less performant virtual machine setup than last time.

robertsLando (approved):

PR #1037, opened 2025-07-12
feat: automate benchmark on PR requests using Github Actions

seriousme:
This PR adds Github actions to automate benchmark running on PR requests

It will create a [Github Job Summary](https://github.blog/news-insights/product-news/supercharging-github-actions-with-job-summaries/) of the benchmark and add it to the benchmark action (see https://github.com/moscajs/aedes/actions/runs/16239498783#summary-45853979456 )

The benchmark steps run serialized in the hope that performance on a single runner does not vary too much.
I have been building testing this using https://github.com/seriousme/aedes/pull/1 which also shows how it operates.

It also shows that variation between results is inevitable, even when running on the same runner and the Aedes code is identical !!

In https://github.com/seriousme/aedes/pull/1  I was able to add the comment automatically to the PR using  `.github/actions/sticky-pr-comment/action.yml`  however if you send a PR from a fork the GITHUB_TOKEN that runs the actions by default (event: [pull_request](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request)  ) only has `read` access on the PR for security reasons (see [Pull Request events for forked repositories](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull-request-events-for-forked-repositories))

This can be fixed by attaching to the event: [pull_request_target](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target), however Github notes explicitly: *Avoid using this event if you need to build or run code from the pull request.*

So for now I used Github Job Summary instead.

Kind regards,
Hans

### DIFF
--- a/.github/workflows/benchmark-compare-serial.yml
+++ b/.github/workflows/benchmark-compare-serial.yml
@@ -56,6 +56,5 @@ jobs:
         shell: bash
         run: |
           cat results.main.csv results.branch.csv | node main-source/benchmarks/report.js >> $GITHUB_STEP_SUMMARY
-          exit $?
 
      
--- a/.github/workflows/benchmark-compare-serial.yml
+++ b/.github/workflows/benchmark-compare-serial.yml
@@ -5,24 +5,20 @@ description: "Runs benchmarks and posts or updates a sticky PR comment with the 
 on:
   pull_request:
     types: [opened, synchronize, reopened]
+  workflow_dispatch:
+    inputs:
+      node-version:
+        description: "Node.js version"
+        default: "20"
+        required: false
+      main-ref:
+        description: "Git ref for the main branch (e.g. main or master)"
+        default: "main"
+        required: false
 
 permissions:
   contents: read
   pull-requests: write # Required for commenting on PRs
-
-inputs:
-  node-version:
-    description: "Node.js version"
-    default: "20"
-    required: false
-  main-ref:
-    description: "Git ref for the main branch (e.g. main or master)"
-    default: "main"
-    required: false
-  report-tag:
-    description: "Tag added to the PR comment to identify it for updating"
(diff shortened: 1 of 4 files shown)


.github/workflows/benchmark-compare-serial.yml
```diff
@@ -0,0 +1,65 @@
...
+      - name: Checkout main branch
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ env.mainRef }}
+          path: main-source
+
+      - name: Set up Node.js (main)
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ env.nodeVersion }}
```
robertsLando, line 30: cache?

.github/workflows/benchmark-compare-serial.yml
```diff
@@ -0,0 +1,65 @@
...
+      # Run benchmark on PR branch
+      - name: Checkout PR branch
+        uses: actions/checkout@v4
+        with:
+          path: pr-source
+
+      - name: Set up Node.js (PR)
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ env.nodeVersion }}
```
robertsLando, line 49: cache?

robertsLando:
@seriousme I'm wondering also about adding an overall `%` for all tests, an average of the results, like -3%.

I think that we could consider there is a performance impact when we see performances below 10%

robertsLando:
@seriousme Will do the major bump once we merge this one :)

seriousme:
Do you also want to migrate to ESM before the major?

.github/workflows/benchmark-compare-serial.yml
seriousme: done

seriousme:
`npm: cache`  only works if `package-lock.json` is present

Issue #1038, opened 2025-07-15
feat: migrate to ESM


robertsLando:
I use to use this as a checklist: https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c 😄

PR #1039, opened 2025-07-15
chore: migrate to ESM

seriousme:
This PR migrates Aedes to ESM.
I followed Sindre Sorhus' checklist mentioned by Daniel in #1038:

- [X] Add "type": "module" to your package.json 
- [X] Replace "main": "index.js" with "exports": "./index.js" in your package.json. 
- [X] Update the "engines" field in package.json to Node.js 20: "node": ">=20".
- [X] Remove 'use strict'; from all JavaScript files.
- [X] Replace all require()/module.export with import/export.
- [X] Use only full relative file paths for imports: import x from '.'; → import x from './index.js';.
- [X] If you have a TypeScript type definition (for example, index.d.ts), update it to use ESM imports/exports.
- [X] Use the [node: protocol](https://nodejs.org/api/esm.html#esm_node_imports) for Node.js built-in imports.

And added the following items:
- [X] remove `proxyquire` as it does not work with ESM (Proxy-require, the name says it all)
- [X] monkey patch neostandard in eslint.config.js as it does not support import assertions yet (I opened a ticket with them, but this works)
- [X] add `test/require.cjs` which checks if `require("aedes")` still works for those that need it.

Enjoy!
Hans
For reference: I used `ts2esm` with a simple `jsconfig.json` to do most of the heavy lifting.

### DIFF
--- a/test/require.cjs
+++ b/test/require.cjs
@@ -1,0 +1,24 @@
+// a test to see if CJS require still works
+
+const { Aedes } = require('../aedes.js')
+const defaultExport = require('../aedes.js')
+const { test } = require('tap')
+
+test('test Aedes constructor', function (t) {
+  t.plan(1)
+  const aedes = new Aedes()
+  t.equal(aedes instanceof Aedes, true, 'Aedes constructor works')
+})
+
+test('test warning on default export', function (t) {
+  t.plan(1)
+  t.throws(defaultExport, 'received expected error')
+})
+
+test('test aedes.createBroker', (t) => {
+  t.plan(1)
+  Aedes.createBroker().then((broker) => {
+    t.teardown(broker.close.bind(broker))
+    t.pass('Aedes.createBroker works')
+  })
+})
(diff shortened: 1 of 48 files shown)


test/require.cjs
```diff
@@ -0,0 +1,24 @@
+// a test to see if CJS require still works
+
+const { Aedes } = require('../aedes.js')
+const defaultExport = require('../aedes.js')
+const { test } = require('tap')
+
+test('test Aedes constructor', function (t) {
```
robertsLando, line 7: was going to ask for this, well done :) Don't know if this still requires the experimental flag btw?

robertsLando (commented):
Everything looks good to me, well done! 🙏🏼 

Let me check if I understand why tests are failing

mcollina (approved):
lgtm

robertsLando (approved):

robertsLando:
@seriousme anything else missing or can I go for the new major? 🚀

seriousme:
I got a dependency update that is almost done in the works that will bring all deps to their latest version, (including Tap to v21 and @sinonjs/fake-timers to v14)  but I can only finish it tonight (after my daytime job :wink:)

PR #1041, opened 2025-07-16
chore: update dependencies

seriousme:
This PR updates all dependencies to their latest versions:
 - @sinonjs/fake-timers   ^11.2.2  →   ^14.0.0
 - @types/node           ^24.0.12  →  ^24.0.14
 - eslint                 ^9.30.1  →   ^9.31.0
 - mqemitter               ^7.0.0  →    ^7.1.0
 - release-it             ^19.0.3  →   ^19.0.4
 - tap                   ^16.3.10  →   ^21.1.0

It removes the following dev dependencies that were no longer used:
- typescript
- snazzy

The upgrade of tap required:
- a new .taprc as the contents were not relevant anymore for v21
- adjustment of package.json as e.g. the  Tap -J flag is now default behavior
- fixing of a few tests in `client-pub-sub.js ` that incorrectly passed too many arguments to t.pass()
- adjustment of CI.yml because the location of the coverage folder was changed by Tap

It also removes a console.log() from `test/connect.js` which was a leftover from debugging. 
 
As a bonus I modified CI.yml slightly so that the first job of the matrix will always upload coverage and we do not need to update the nodeJS version in this specific "if"  anymore.

Kind regards,
Hans


seriousme:
Btw: benchmark results are within 4% tolerance 😉 
https://github.com/moscajs/aedes/actions/runs/16326160437

robertsLando:
> Btw: benchmark results are within 4% tolerance 😉

Yep I saw that :) Good!

> This all within the same GitHub runner!

Sincerly I have no clue why there is that much difference, maybe @mcollina could have some hints about that

.github/workflows/ci.yml
```diff
@@ -61,10 +61,11 @@ jobs:
           npm run test:ci
       
       - name: Upload coverage reports to Codecov
-        if: matrix.node-version == 20 && matrix.os == 'ubuntu-latest'
+        # only let the first job upload coverage info
+        if: strategy.job-index == 0
```
robertsLando, line 65: this is new to me, good to know :)

robertsLando (approved):
LGTM. Release after this? We may also drop tap in a future PR in favor of node tests

seriousme:
> We may also drop tap in a future PR in favor of node tests

It is on my todo 😉

robertsLando:
> I saw some projects use AWS EC2 Baremetal instances to avoid this, but those come at a price.

Will see if I find some free tier on some cloud provider that we could use for this :)

PR #1042, opened 2025-07-26
chore: replace tap by node:test

seriousme:
This PR replaces Tap by node:test

Tasks:
-  [x] auth.js
-  [x] basic.js
-  [x] bridge.js
-  [x] client-pub-sub.js
-  [x] close_socket_by_other_party.js
-  [x] connect.js  (with TODO's)
-  [x] events.js
-  [x] keep-alive.js
-  [x] meta.js (with 1 TODO)
-  [x] not-blocking.js
-  [x] qos1.js
-  [x] qos2.js
-  [x] regr-21.js
-  [x] require.cjs
-  [x] retain.js (with 1 TODO)
-  [x] topics.js
-  [x] will.js
-  [x] update package.json scripts
-  [x] update package.json devDependencies
-  [x] update 'coverage' directory in ci.yml

The following devDependencies have been replaced by NodeJS native functionality
 - @sinonjs/fake-timers
 - concat-stream
 - duplexify
 - mqtt-connection
 - tap

It adds `c8` as devDependency to facilitate test coverage.

### DIFF
--- a/test/helper.js
+++ b/test/helper.js
@@ -253,10 +253,18 @@ export async function withTimeout (promise, timeoutMs, timeoutResult) {
 }
 
 /**
- * @param {Object} parser - The emitter object
- * @param {Readable} sourceStream - The source stream to read MQTT packets from
- * @param {Object} opts - low and high water mark
- * @returns {AsyncGenerator} An async generator that yields MQTT packets
+ * Asynchronously yields MQTT packets parsed from a source stream, with backpressure management.
+ *
+ * Backpressure is controlled using a queue and the `highWaterMark`/`lowWaterMark` parameters:
+ * - When the number of buffered packets in the queue reaches `highWaterMark`, the source stream is paused to prevent overload.
+ * - When the queue size drops to or below `lowWaterMark`, the source stream is resumed to allow more data to flow.
+ *
+ * @param {Object} parser - The emitter object that parses MQTT packets and emits 'packet' events.
+ * @param {Readable} sourceStream - The source stream to read MQTT packets from.
+ * @param {Object} [opts] - Options for backpressure control.
+ * @param {number} [opts.highWaterMark=2] - Maximum number of packets to buffer before pausing the source stream.
+ * @param {number} [opts.lowWaterMark=0] - Minimum number of packets in the buffer to resume the source stream.
+ * @returns {AsyncGenerator<Object|null, void, unknown>} An async generator that yields MQTT packets, or null when done.
  */
 async function * packetGenerator (parser, sourceStream, opts = {
   highWaterMark: 2,
--- a/test/will.js
+++ b/test/will.js
@@ -55,7 +55,7 @@ test('delivers a will', async (t) => {
 
   const { packet } = await oneWillFromBroker(s.broker)
   t.assert.equal(packet.topic, opts.will.topic, 'topic matches')
-  t.assert.deepEqual(structuredClone(packet).payload, opts.will.payload, 'payload matches')
+  t.assert.deepEqual(packet.payload, opts.will.payload, 'payload matches')
   t.assert.equal(packet.qos, opts.will.qos, 'qos matches')
   t.assert.equal(packet.retain, opts.will.retain, 'retain matches')
 })
@@ -109,7 +109,7 @@ test('delivers old will in case of a crash', async (t) => {
   const { packet, received } = await oneWillFromBroker(broker)
   t.assert.ok(Date.now() - start >= 3 * interval, 'the will needs to be emitted after 3 heartbeats')
   t.assert.equal(packet.topic, will.topic, 'topic matches')
-  t.assert.deepEqual(structuredClone(packet).payload, will.payload, 'payload matches')
+  t.assert.deepEqual(packet.payload, will.payload, 'payload matches')
   t.assert.equal(packet.qos, will.qos, 'qos matches')
   t.assert.equal(packet.retain, will.retain, 'retain matches')
   t.assert.equal(authorized, true, 'authorization called')


seriousme:
FYI:  the original bridge.js has a problem, it ignores test failures of the second test:

https://github.com/moscajs/aedes/blob/b5178c02629d8e2096820c4b743a591b26a49fda/test/bridge.js#L40-L46

because of a lack of `test.plan()` and test assertions, which makes the test fail silently.

This test always fails because `mqtt-packet/writeToStream.js` does not support the bridge protocol version `128+4`
https://github.com/mqttjs/mqtt-packet/blob/e39fb28c10628720fb50e5eb355492684ff0caaa/writeToStream.js#L108

For now I have converted the second test in bridge.js to work with `node:test` but commented it out.
We could create a pre-encoded packet and send that to the connection, but the better way imo would be to add the bridge protocol versions to writeToStream.js

Kind regards,
Hans

robertsLando (commented):
Well done so far! :)

seriousme:
> > We could create a pre-encoded packet and send that to the connection, but the better way imo would be to add the bridge protocol versions to writeToStream.js
> 
> I suggest the latter 🙏🏼

see PR on mqtt-packet 😉

seriousme:
btw: I have an idea to solve the not-blocking.js on Windows and Mac, will try to fix this myself

test/connect.js
```diff
@@ -1,522 +1,450 @@
...
   t.plan(3)
 
-  Aedes.createBroker().then((broker) => {
-    t.teardown(broker.close.bind(broker))
+  const broker = await Aedes.createBroker()
+  t.after(() => broker.close())
+  const s = setup(broker)
 
-    const s = setup(broker)
+  // s.inStream.write throws an error when trying to write illegal packets
```
robertsLando, line 122: wondering why this didn't throw before?

test/connect.js
```diff
@@ -1,522 +1,450 @@
...
+    expectedReturnCode: 6
   })
+
+  const [client, err] = await once(s.broker, 'connectionError')
+  t.assert.ok(client, 'client is defined')
+  t.assert.equal(err.message, 'keep alive limit exceeded')
+  t.assert.equal(s.broker.connectedClients, 0)
 })
 
+// TODO: test fails because Aedes does not reject this
```
robertsLando, line 112: is this due to changes on aedes code or race conditions? It's not clear by checking the test

robertsLando (changes requested):

seriousme:
not-blocking works again.

test/connect.js
seriousme: I think that the s.instream.write error was previously absorbed by the callbacks. The test succeeded because the client was expected to not connect and it did not connect. The reason for not connecting was however not a rejection because of a missing clientid but because the connect is never sent.

test/connect.js
seriousme: I think this was the same as the other one.

test/connect.js
robertsLando: yeah sometimes it's hard to understand. BTW everything looks good for now 🙏🏼 Well done

robertsLando:
@seriousme is this ready for full review?

test/helperAsync.js
```diff
@@ -0,0 +1,358 @@
...
+
+  const onPacket = packet => {
+    if (waiting) {
+      waiting(packet)
+      waiting = null
+    } else {
+      queue.push(packet)
+    }
+    // Pause source if queue is full
+    if (queue.length > highWaterMark) {
```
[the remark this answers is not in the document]
seriousme: Fixed

seriousme:
> FYI I will be on vacation starting from next week till the end of August 🙏🏼
Enjoy your vacation! I am not in a rush :wink:

seriousme:
@robertsLando 
FYI: 
All items are in now, we can let go of:

- @sinonjs/fake-timers
- concat-stream
- duplexify
- mqtt-connection
- tap

The TODO in qos2.js has been resolved. It turns out the when the client gets the connack, Aedes might not have finished the connect yet:
```js
  const subscriber2 = setup(broker)
  const connack = await connect(subscriber2, { connect: opts })
  t.assert.equal(connack.sessionPresent, true, 'session present is set to true')
  // getting a connack does not mean that the client setup has completed in Aedes
  await once(broker, 'clientReady')
  await publish(t, publisher, expected)
  await receive(t, subscriber2, expected)
```
I am still investigation some of the other TODO's but for now it is ready for review.

Enjoy your vacation!

robertsLando:
@seriousme Seems `NODE_OPTIONS="--no-experimental-strip-types"` could be an easy fix

seriousme:
> @seriousme Seems `NODE_OPTIONS="--no-experimental-strip-types"` could be an easy fix

I had the same idea and I tried yesterday, but it results in the same error as option 4: `option '--no-experimental-strip-types' is not supported`

test/meta.js
```diff
@@ -1,342 +1,374 @@
...
+  // run parallel
+  await Promise.all([
+    checkUnsubscribe(),
+    doUnsubscribe
+  ])
+})
 
-  Aedes.createBroker().then((broker) => {
-    t.teardown(broker.close.bind(broker))
+// TODO: Aedes does emit an unsubscribe event on client close
```
robertsLando, line 258: TODO

robertsLando (commented):
I added missing TODOs just for reference

robertsLando:
This PR is HUGE 😆 Well done @seriousme , I liked a lot that you dropped mqtt-connection as well replaced with the async generator 💪🏼

test/helper.js
```diff
@@ -1,107 +1,377 @@
...
+        } else {
+          // There is buffered data ready
+          // If we're emptying the queue, resume source
+          if (queue.length <= lowWaterMark) {
+            sourceStream.resume()
+          }
+          yield queue.shift()
+        }
+      }
+    }
```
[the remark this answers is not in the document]
seriousme: The generator ends to exist as soon as the consumer of the generator ends.

test/qos2.js
```diff
@@ -1,701 +1,576 @@
...
-        retain: false,
-        dup: false
-      }, 'pubrel must match')
-
-      if (done) {
-        done()
-      }
-    })
+  const pubRel = await nextPacket(subscriber)
+  t.assert.deepStrictEqual(structuredClone(pubRel), {
```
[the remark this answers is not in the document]
seriousme: Incorrect finding: pubrel is still of class packet

seriousme:
@robertsLando  
Good to read that you are back! Hope you had a good time.

While you were away I also made preparations for an additional PR that reduces the number of prod-dependencies 😉 , but we first need to finalize this PR. I would rather not mix these changes on testing with changes on prod.

What do you want to do with the conversations above that you marked as TODO?
E.g. should we make separate issues for them or a single issue with a tasklist?

As far as I can see I answered all your questions. If I missed one then please let me know!

Kind regards,
Hans

test/will.js
```diff
@@ -10,487 +15,371 @@ async function memorySetup (opts) {
...
-  t.plan(4)
+test('calling close two times should not deliver two wills', async (t) => {
+  t.plan(5)
+
+  const opts = addWillToOpts({})
+  const s = await createAndConnect(t, { connect: opts })
+
+  const received = willsFromBroker(s.broker)
+  s.client.close()
+  s.client.close()
```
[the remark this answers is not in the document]
robertsLando: lol sorry, oversight

seriousme:
> > I think mqtt-packet should treat this as an error
> 
> Do you mean this kind of error should never be thrown by broker? If so the packet parse should emit an error couldn't we intercept that?

I think that low level packet errors should be detected at packet decoding level by the parser and then based on the error thrown by the parser the broker should close the connection.
e.g. The  MQTT standard says that any protocol violation must result in a close of the connection.

That way the broker only sees valid packets and does not need to bother about low level issues.
(seperation of concerns etc)

Kind regards,
Hans

robertsLando:
> That way the broker only sees valid packets and does not need to bother about low level issues.
(seperation of concerns etc)

I agree, so I think that all tests related to this could be removed IMO, so 3-4-6 should be left open, 7 should be resolved on mqtt-packet repo (open an issue there?) and the others could be removed. WDYT?

seriousme:
I think that from this list:

1) connect.js: 'reject clients with no clientId running on MQTT 3.1.0'
2) connect.js: 'reject clients without clientid and clean=false on MQTT 3.1.1'
3) connect.js: 'After first CONNECT Packet, others are queued until 'connect' event'
As far as I know there is no way to halt processing after succesful connect
4) connect.js: 'Test queue limit'
This is the same problem as 3, the only thing we can test is the queue limit of 2 before connect by pausing authenticate.
5) meta.js: 'emit unsubscribe event if unrecognized params in unsubscribe packet structure'
I think that the mqtt-packet parser should refuse unrecognized params
6) meta.js: 'dont emit unsubscribe event on client close'
This seems to be a bug in Aedes
7) mqtt-packet allows incorrect combinations of protocolId and protocolVersion
e.g. : version 3 with id 'MQTT' is accepted, so is version 5 with 'MQIdsp'

Items 1,2,5 and 7 should be fixed in MQTT-packet  to ensure they emit an error and the tests should be removed here
Items 3 and 4 are currently just not testable so the tests should be left open or removed
Item 6 should be left open and investigated more deeply as this is probably a bug.
(next PR should make it easier to trace item 6)

Kind regards,
Hans

seriousme:
@robertsLando do you agree?

PR #1045, opened 2025-09-02
chore: reduce production dependencies

seriousme:
@robertsLando 
As promised, this PR reduces  the number of production dependencies to a bare minimum.

This PR replaces:
1. "end-of-stream" by native "finished" from "node:stream" 
2. "fastfall" by a local  "runfall" 
3. "fastparallel":  by a local "runParallel" 
4. "fastseries": by a local "runSeries" 
5. "hyperid":  by  "randomUUID" from 'node:crypto' , hyperid might be faster (25M ops/second vs 12M ops/second) but also creates predictable ID's and since we only used "hyperid" to generate clientID's  12M ops/second should still be enough imo)
6. "retimer":  by a local  "retimer"  which is nearly as fast as the original "retimer" module with a difference of 1 second on 1M invocations.
7. "reusify": by a local  "ObjectPool"  (btw: reusify was only used on "Enquers" and I am not sure how much performance this actually added as V8 has optimized object creation over time, but just to be sure I put in the ObjectPool)
8. "uuid":  by  "randomUUID" from 'node:crypto' as it turned out that on nodeJS  the "uuidv4" of "uuid"  it is an alias for "randomUUID" from 'node:crypto'  anyway.

The local variants listed above are all quite small, mostly because they only need to support one variant instead of the many variants that some external modules provide. Also over the years some patterns have become more easy to implement with modern Javascript.

This PR tries to keep the code changes to a minimum.
From here on, further optimizations might be possible, but I didn't want to make this PR too complex.

Kind regards,
Hans

### DIFF
--- a/aedes.js
+++ b/aedes.js
@@ -381,9 +381,9 @@ function runSeries (state, actions, packet, done) {
   next()
 }
 
-async function closeClient (client) {
+async function closeClient (clientInstance) {
   return new Promise((resolve) => {
-    client.close(resolve)
+    clientInstance.close(resolve)
   })
 }
 
--- a/aedes.js
+++ b/aedes.js
@@ -5,7 +5,7 @@ import Packet from 'aedes-packet'
 import memory from 'aedes-persistence'
 import mqemitter from 'mqemitter'
 import Client from './lib/client.js'
-import { $SYS_PREFIX, bulk, ObjectPool } from './lib/utils.js'
+import { $SYS_PREFIX, bulk } from './lib/utils.js'
 import pkg from './package.json' with { type: 'json' }
 
 const defaultOptions = {
@@ -54,7 +54,6 @@ export class Aedes extends EventEmitter {
     }
 
     this._series = runSeries
-    this._enqueuers = new ObjectPool(DoEnqueues)
 
     this.preConnect = opts.preConnect
     this.authenticate = opts.authenticate
@@ -297,7 +296,7 @@ function emitPacket (packet, done) {
 }
 
 function enqueueOffline (packet, done) {
-  const enqueuer = this.broker._enqueuers.get()
+  const enqueuer = new DoEnqueues()
 
   enqueuer.complete = done
   enqueuer.packet = packet
@@ -341,7 +340,6 @@ class DoEnqueues {
 
       broker.persistence.outgoingEnqueueCombi(subs, packet)
         .then(() => complete(null), complete)
-      broker._enqueuers.release(that)
     }
   }
 }
--- a/aedes.js
+++ b/aedes.js
@@ -5,7 +5,7 @@ import Packet from 'aedes-packet'
 import memory from 'aedes-persistence'
 import mqemitter from 'mqemitter'
 import Client from './lib/client.js'
-import { $SYS_PREFIX, batch } from './lib/utils.js'
+import { $SYS_PREFIX, batch, noop, runSeries } from './lib/utils.js'
 import pkg from './package.json' with { type: 'json' }
 
 const defaultOptions = {
@@ -352,17 +352,6 @@ const publishFuncsQoS = [
   callPublished
 ]
 
-// runSeries runs functions in fastseries style
-function runSeries (state, actions, packet, done) {
-  done = (done || noop).bind(state)
-  let i = 0
-  function next (err) {
-    if (err || i === actions.length) return done(err)
-    actions[i++].call(state, packet, next)
-  }
-  next()
-}
-
 async function closeClient (clientInstance) {
   return new Promise((resolve) => {
     clientInstance.close(resolve)
@@ -403,8 +392,6 @@ class PublishState {
     this.packet = packet
   }
 }
-
-function noop () { }
 
 function warnMigrate () {
   throw new Error(
--- a/aedes.js
+++ b/aedes.js
@@ -1,11 +1,11 @@
 import EventEmitter from 'node:events'
 import { randomUUID } from 'node:crypto'
-import { pipeline } from 'stream'
+import { promisify } from 'node:util'
 import Packet from 'aedes-packet'
 import memory from 'aedes-persistence'
 import mqemitter from 'mqemitter'
 import Client from './lib/client.js'
-import { $SYS_PREFIX, bulk } from './lib/utils.js'
+import { $SYS_PREFIX, batch } from './lib/utils.js'
 import pkg from './package.json' with { type: 'json' }
 
 const defaultOptions = {
@@ -106,61 +106,45 @@ export class Aedes extends EventEmitter {
       }, noop)
     }
 
-    function deleteOldBrokers (broker) {
-      if (that.brokers[broker] + (3 * opts.heartbeatInterval) < Date.now()) {
-        delete that.brokers[broker]
+    async function _clearWills () {
+      const pAuthorizePublish = promisify(that.authorizePublish).bind(that)
+      const pPublish = promisify(that.publish).bind(that)
+      const batchSize = 16 // default highWatermark for Writable in ObjectMode
+
+      async function checkAndPublish (will) {
+        const notPublish = that.brokers[will.brokerId] !== undefined && that.brokers[will.brokerId] + (3 * opts.heartbeatInterval) >= Date.now()
+        if (notPublish) {
+          return
+        }
+        // randomize this, so that multiple brokers
+        // do not publish the same wills at the same time
+        const client = that.clients[will.clientId] || null
+        await pAuthorizePublish(client, will)
+        await pPublish(will)
+        await that.persistence.delWill({
+          id: will.clientId,
+          brokerId: will.brokerId
+        })
+      }
+
+      // delete old brokers
+      for (const broker in that.brokers) {
+        if (that.brokers[broker] + (3 * opts.heartbeatInterval) < Date.now()) {
+          delete that.brokers[broker]
+        }
+      }
+
+      const wills = that.persistence.streamWill(that.brokers)
+      for await (const promises of batch(wills, checkAndPublish, batchSize)) {
+        await Promise.all(promises)
       }
     }
 
-    this._clearWillInterval = setInterval(function () {
-      Object.keys(that.brokers).forEach(deleteOldBrokers)
-
-      pipeline(
-        that.persistence.streamWill(that.brokers),
-        bulk(receiveWills),
-        function done (err) {
-          if (err) {
-            that.emit('error', err)
-          }
-        }
-      )
-    }, opts.heartbeatInterval * 4)
-
-    async function receiveWills (chunks, done) {
-      try {
-        await Promise.all(chunks.map(chunk => new Promise((resolve, reject) => {
-          checkAndPublish(chunk, (err) => {
-            if (err) reject(err)
-            else resolve()
-          })
-        })))
-        done()
-      } catch (err) {
-        done(err)
-      }
-    }
-
-    function checkAndPublish (will, done) {
-      const notPublish = that.brokers[will.brokerId] !== undefined && that.brokers[will.brokerId] + (3 * opts.heartbeatInterval) >= Date.now()
-
-      if (notPublish) return done()
-
-      // randomize this, so that multiple brokers
-      // do not publish the same wills at the same time
-      that.authorizePublish(that.clients[will.clientId] || null, will, function (err) {
-        if (err) { return doneWill() }
-        that.publish(will, doneWill)
-
-        function doneWill (err) {
-          if (err) { return done(err) }
-          that.persistence.delWill({
-            id: will.clientId,
-            brokerId: will.brokerId
-          }).then(will => done(undefined, will), done)
-        }
+    this._clearWillInterval = setInterval(() => {
+      _clearWills().catch(err => {
+        that.emit('error', err)
       })
-    }
-
+    }, opts.heartbeatInterval * 4)
     this.mq.on($SYS_PREFIX + '+/heartbeat', function storeBroker (packet, done) {
       that.brokers[packet.payload.toString()] = Date.now()
       done()
--- a/aedes.js
+++ b/aedes.js
@@ -258,7 +258,7 @@ export class Aedes extends EventEmitter {
     for (const clientId of Object.keys(this.clients)) {
       promises.push(closeClient(this.clients[clientId]))
     }
-    Promise.all(promises).then(() => {
+    Promise.all(promises).finally(() => {
       that.emit('closed')
       that.mq.close(cb)
     })
--- a/aedes.js
+++ b/aedes.js
@@ -1,14 +1,11 @@
 import EventEmitter from 'node:events'
-import parallel from 'fastparallel'
-import series from 'fastseries'
-import { v4 as uuidv4 } from 'uuid'
-import reusify from 'reusify'
+import { randomUUID } from 'node:crypto'
 import { pipeline } from 'stream'
 import Packet from 'aedes-packet'
 import memory from 'aedes-persistence'
 import mqemitter from 'mqemitter'
 import Client from './lib/client.js'
-import { $SYS_PREFIX, bulk } from './lib/utils.js'
+import { $SYS_PREFIX, bulk, ObjectPool } from './lib/utils.js'
 import pkg from './package.json' with { type: 'json' }
 
 const defaultOptions = {
@@ -37,7 +34,7 @@ export class Aedes extends EventEmitter {
 
     opts = Object.assign({}, defaultOptions, opts)
     this.opts = opts
-    this.id = opts.id || uuidv4()
+    this.id = opts.id || randomUUID()
     // +1 when construct a new aedes-packet
     // internal track for last brokerCounter
     this.counter = 0
@@ -56,9 +53,8 @@ export class Aedes extends EventEmitter {
       return new Client(that, conn, req)
     }
 
-    this._parallel = parallel()
-    this._series = series()
-    this._enqueuers = reusify(DoEnqueues)
+    this._series = runSeries
+    this._enqueuers = new ObjectPool(DoEnqueues)
 
     this.preConnect = opts.preConnect
     this.authenticate = opts.authenticate
@@ -131,8 +127,18 @@ export class Aedes extends EventEmitter {
       )
     }, opts.heartbeatInterval * 4)
 
-    function receiveWills (chunks, done) {
-      that._parallel(that, checkAndPublish, chunks, done)
+    async function receiveWills (chunks, done) {
+      try {
+        await Promise.all(chunks.map(chunk => new Promise((resolve, reject) => {
+          checkAndPublish(chunk, (err) => {
+            if (err) reject(err)
+            else resolve()
+          })
+        })))
+        done()
+      } catch (err) {
+        done(err)
+      }
     }
 
     function checkAndPublish (will, done) {
@@ -142,7 +148,7 @@ export class Aedes extends EventEmitter {
 
       // randomize this, so that multiple brokers
       // do not publish the same wills at the same time
-      this.authorizePublish(that.clients[will.clientId] || null, will, function (err) {
+      that.authorizePublish(that.clients[will.clientId] || null, will, function (err) {
         if (err) { return doneWill() }
         that.publish(will, doneWill)
 
@@ -265,11 +271,14 @@ export class Aedes extends EventEmitter {
     this.closed = true
     clearInterval(this._heartbeatInterval)
     clearInterval(this._clearWillInterval)
-    this._parallel(this, closeClient, Object.keys(this.clients), doneClose)
-    function doneClose () {
+    const promises = []
+    for (const clientId of Object.keys(this.clients)) {
+      promises.push(closeClient(this.clients[clientId]))
+    }
+    Promise.all(promises).then(() => {
       that.emit('closed')
       that.mq.close(cb)
-    }
+    })
   }
 }
 
@@ -361,8 +370,21 @@ const publishFuncsQoS = [
   callPublished
 ]
 
-function closeClient (client, cb) {
-  this.clients[client].close(cb)
+// runSeries runs functions in fastseries style
+function runSeries (state, actions, packet, done) {
+  done = (done || noop).bind(state)
+  let i = 0
+  function next (err) {
+    if (err || i === actions.length) return done(err)
+    actions[i++].call(state, packet, next)
+  }
+  next()
+}
+
+async function closeClient (client) {
+  return new Promise((resolve) => {
+    client.close(resolve)
+  })
 }
 
 function defaultPreConnect (client, packet, callback) {
@@ -400,11 +422,11 @@ class PublishState {
   }
 }
 
-function noop () {}
+function noop () { }
 
 function warnMigrate () {
   throw new Error(
-` Aedes default export has been removed.
+    ` Aedes default export has been removed.
  Use 'const aedes = await Aedes.createBroker()' instead.
  See: https://github.com/moscajs/aedes/docs/MIGRATION.MD
  `)
--- a/aedes.js
+++ b/aedes.js
@@ -54,8 +54,6 @@ export class Aedes extends EventEmitter {
       return new Client(that, conn, req)
     }
 
-    this._series = runSeries
-
     this.preConnect = opts.preConnect
     this.authenticate = opts.authenticate
     this.authorizePublish = opts.authorizePublish
@@ -200,7 +198,7 @@ export class Aedes extends EventEmitter {
     const p = new Packet(packet, this)
     const publishFuncs = p.qos > 0 ? publishFuncsQoS : publishFuncsSimple
 
-    this._series(new PublishState(this, client, packet), publishFuncs, p, done)
+    runSeries(new PublishState(this, client, packet), publishFuncs, p, done)
   }
 
   subscribe (topic, func, done) {
--- a/lib/client.js
+++ b/lib/client.js
@@ -1,14 +1,13 @@
 import mqtt from 'mqtt-packet'
 import EventEmitter from 'node:events'
 import util from 'util'
-import eos from 'end-of-stream'
 import Packet from 'aedes-packet'
 import write from './write.js'
 import QoSPacket from './qos-packet.js'
 import handleSubscribe from './handlers/subscribe.js'
 import handleUnsubscribe from './handlers/unsubscribe.js'
 import handle from './handlers/index.js'
-import { pipeline } from 'stream'
+import { pipeline, finished } from 'stream'
 import { through } from './utils.js'
 
 class Client {
@@ -84,7 +83,7 @@ class Client {
     this._parser.on('error', this.emit.bind(this, 'error'))
 
     conn.on('end', this.close.bind(this))
-    this._eos = eos(this.conn, this.close.bind(this))
+    this._eos = finished(this.conn, this.close.bind(this))
 
     const getToForwardPacket = (_packet) => {
       // Mqttv5 3.8.3.1: https://docs.oasis-open.org/mqtt/mqtt/v5.0/mqtt-v5.0.html#_Toc3901169
--- a/lib/client.js
+++ b/lib/client.js
@@ -233,7 +233,7 @@ class Client {
     this._parser._queue = null
 
     if (this._keepaliveTimer) {
-      this._keepaliveTimer.clear()
+      clearTimeout(this._keepaliveTimer)
       this._keepaliveInterval = -1
       this._keepaliveTimer = null
     }
--- a/lib/client.js
+++ b/lib/client.js
@@ -170,9 +170,9 @@ class Client {
     this.conn.removeAllListeners('error')
     this.conn.on('error', noop)
     // hack to clean up the write callbacks in case of error
-    const state = this.conn._writableState
-    const list = typeof state.getBuffer === 'function' ? state.getBuffer() : state.buffer
-    list.forEach(drainRequest)
+    if (!this.conn.destroyed) {
+      this.conn.destroy(err)
+    }
     this.broker.emit(this.id ? 'clientError' : 'connectionError', this, err)
     this.close()
   }
@@ -458,10 +458,6 @@ function writeQoS (err, client, packet) {
       packet.writeCallback()
     })
   }
-}
-
-function drainRequest (req) {
-  req.callback()
 }
 
 util.inherits(Client, EventEmitter)
--- a/lib/handlers/connect.js
+++ b/lib/handlers/connect.js
@@ -1,12 +1,9 @@
-import retimer from 'retimer'
+import { randomUUID } from 'node:crypto'
 import { pipeline } from 'stream'
 import write from '../write.js'
 import QoSPacket from '../qos-packet.js'
-import { through } from '../utils.js'
+import { through, retimer } from '../utils.js'
 import handleSubscribe from './subscribe.js'
-import hyperid from 'hyperid'
-
-const uniqueId = hyperid()
 
 class Connack {
   constructor (arg) {
@@ -84,7 +81,7 @@ function init (client, packet, done) {
     return
   }
 
-  client.id = clientId || 'aedes_' + uniqueId()
+  client.id = clientId || 'aedes_' + randomUUID()
   client.clean = packet.clean
   client.version = packet.protocolVersion
   client._will = packet.will
--- a/lib/handlers/connect.js
+++ b/lib/handlers/connect.js
@@ -2,7 +2,7 @@ import { randomUUID } from 'node:crypto'
 import { pipeline } from 'stream'
 import write from '../write.js'
 import QoSPacket from '../qos-packet.js'
-import { through, retimer } from '../utils.js'
+import { through } from '../utils.js'
 import handleSubscribe from './subscribe.js'
 
 class Connack {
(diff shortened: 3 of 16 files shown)


seriousme:
Automated Benchmark testing suggests more than 10℅  perf decrease. 
I will try to figure out what is causing this.

Kind regards,
Hans

seriousme:
It urns out the difference was actually positive instead of negative:

----

 # Overall Benchmark Results

 +x% is better, -x% is worse, current threshold to fail at -10%


| Label | Benchmark | Config | Average | Units | Percentage
|-------|-----------|--------|---------|-------|-----------
| main | sender.js | QoS=0, Cores=4 | 107096 | msg/s | 100%
| reduce-production-dependencies | sender.js | QoS=0, Cores=4 | 124161 | msg/s | +15.93%
| main | receiver.js | QoS=0, Cores=4 | 108229 | msg/s | 100%
| reduce-production-dependencies | receiver.js | QoS=0, Cores=4 | 124055 | msg/s | +14.62%
| main | sender.js | QoS=1, Cores=4 | 53724 | msg/s | 100%
| reduce-production-dependencies | sender.js | QoS=1, Cores=4 | 56895 | msg/s | +5.90%
| main | receiver.js | QoS=1, Cores=4 | 53727 | msg/s | 100%
| reduce-production-dependencies | receiver.js | QoS=1, Cores=4 | 56893 | msg/s | +5.89%
| main | pingpong.js | QoS=1, Cores=4, Score='perc95' | 41 | ms | 100%
| reduce-production-dependencies | pingpong.js | QoS=1, Cores=4, Score='perc95' | 41 | ms | 100%

 # Benchmark Results for main
|Benchmark | Config | Units | Round 1 |Round 2 |Round 3 |Round 4 |Round 5 |Round 6 |Round 7 |Round 8 |Round 9 |Round 10
|----------|--------|-------|---|---|---|---|---|---|---|---|---|---
| sender.js | QoS=0, Cores=4 | msg/s| 104709 |102963 |100781 |116193 |107521 |111121 |108203 |98121 |126638 |94705
| receiver.js | QoS=0, Cores=4 | msg/s| 113003 |105290 |108320 |102153 |100046 |112616 |112741 |107542 |108220 |112359
| sender.js | QoS=1, Cores=4 | msg/s| 50461 |53588 |54380 |55006 |52976 |55709 |54517 |55805 |51788 |53013
| receiver.js | QoS=1, Cores=4 | msg/s| 50489 |53579 |54377 |55014 |52977 |55703 |54515 |55809 |51784 |53020
| pingpong.js | QoS=1, Cores=4, Score='perc95' | ms| 41 |41 |41 |41 |41 |41 |41 |41 |41 |41

 # Benchmark Results for reduce-production-dependencies
|Benchmark | Config | Units | Round 1 |Round 2 |Round 3 |Round 4 |Round 5 |Round 6 |Round 7 |Round 8 |Round 9 |Round 10
|----------|--------|-------|---|---|---|---|---|---|---|---|---|---
| sender.js | QoS=0, Cores=4 | msg/s| 116256 |119349 |122598 |124305 |135292 |132318 |116559 |117091 |120768 |137072
| receiver.js | QoS=0, Cores=4 | msg/s| 124809 |116053 |118338 |131602 |118724 |137263 |133331 |119014 |119005 |122410
| sender.js | QoS=1, Cores=4 | msg/s| 56878 |57472 |56176 |57224 |56130 |56366 |56299 |56725 |58094 |57588
| receiver.js | QoS=1, Cores=4 | msg/s| 56853 |57509 |56168 |57195 |56178 |56317 |56341 |56680 |58099 |57590
| pingpong.js | QoS=1, Cores=4, Score='perc95' | ms| 41 |41 |41 |41 |41 |41 |41 |41 |41 |41

robertsLando (approved):
Love this! ❤️ 

Let's see if @mcollina has some opinions on this otherwise this is good to merge 🙏🏼

aedes.js
```diff
@@ -131,8 +127,18 @@ export class Aedes extends EventEmitter {
...
+        await Promise.all(chunks.map(chunk => new Promise((resolve, reject) => {
+          checkAndPublish(chunk, (err) => {
+            if (err) reject(err)
+            else resolve()
+          })
+        })))
+        done()
+      } catch (err) {
+        done(err)
+      }
```
mcollina, line 141: This is by far not equivalent. It also creates a bazillion of promises and it will significantly impact performance

lib/utils.js
```diff
@@ -37,6 +37,40 @@ export function through (transform) {
...
+  return {
+    reschedule (newMs) {
+      clearTimeout(timeout)
+      timeout = setTimeout(fn, newMs)
+    },
+    clear () {
+      clearTimeout(timeout)
+    }
+  }
+}
```
mcollina, line 72: Do this instead https://nodejs.org/docs/latest/api/timers.html#timeoutrefresh.

lib/handlers/connect.js
```diff
@@ -84,7 +81,7 @@ function init (client, packet, done) {
     return
   }
 
-  client.id = clientId || 'aedes_' + uniqueId()
+  client.id = clientId || 'aedes_' + randomUUID()
```
mcollina, line 84: +1

lib/client.js
```diff
@@ -1,14 +1,13 @@
...
 import util from 'util'
-import eos from 'end-of-stream'
 import Packet from 'aedes-packet'
 import write from './write.js'
 import QoSPacket from './qos-packet.js'
 import handleSubscribe from './handlers/subscribe.js'
 import handleUnsubscribe from './handlers/unsubscribe.js'
 import handle from './handlers/index.js'
-import { pipeline } from 'stream'
+import { pipeline, finished } from 'stream'
```
mcollina, line 10: +1

lib/utils.js
```diff
@@ -37,6 +37,40 @@ export function through (transform) {
...
+
+  get () {
+    return this.pool.pop() || new this.ClassCtor()
+  }
+
+  release (obj) {
+    // Optionally reset the object here if needed
+    this.pool.push(obj)
+  }
+}
```
mcollina, line 55: Doing this would likely be worse that reallocating them all the time.

aedes.js
```diff
@@ -56,9 +53,8 @@ export class Aedes extends EventEmitter {
       return new Client(that, conn, req)
     }
 
-    this._parallel = parallel()
-    this._series = series()
-    this._enqueuers = reusify(DoEnqueues)
+    this._series = runSeries
+    this._enqueuers = new ObjectPool(DoEnqueues)
```
mcollina, line 57: It might be interesting to verify if this is needed at all. My 2 cents is that we don't need this at all anymore and there is no perf gap in just allocating them all the time.

mcollina (changes requested):
I generically don't see a problem with dependencies and number of things on disk, but some of the changes are incorrect or will worsen the performance.

Some of it is in the right direction because it will now be using things from the platform.

aedes.js
```diff
@@ -131,8 +127,18 @@ export class Aedes extends EventEmitter {
...
-      that._parallel(that, checkAndPublish, chunks, done)
+    async function receiveWills (chunks, done) {
+      try {
+        await Promise.all(chunks.map(chunk => new Promise((resolve, reject) => {
+          checkAndPublish(chunk, (err) => {
+            if (err) reject(err)
+            else resolve()
+          })
+        })))
+        done()
```
mcollina, line 138:
```suggestion
        process.nextTick(done)
```

otherwise error handling would be a disaster

aedes.js
seriousme: Just ran the benchmark on the commit that only removes the object pool. No material difference, still about 18% improvement over the main branch. See: https://github.com/moscajs/aedes/actions/runs/17486480074/attempts/1#summary-49666422333

lib/utils.js
seriousme: applied

lib/handlers/subscribe.js
```diff
@@ -1,14 +1,32 @@
...
+          cb.call(ctx, err, nextarg)
+        }
+        return
+      }
+      const fn = fns[i++]
+      fn.call(ctx, nextarg, next)
+    }
+    next(null, arg)
+  }
+}
```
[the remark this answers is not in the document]
seriousme: Not sure what you mean with this comment. The coverage report says that its the `runFallall` is called 36 times and all code of `runfall` is covered during tests. The anonymous inner function runs 164 times, the `next()` function runs 628 times. Do you want me to move it to lib/utils.js and test it standalone?

seriousme:
@robertsLando 
I also added a test to test/wills.js to: 
- show that the new clearwills mechanism works with a lager number of wills
- increase test coverage

lib/handlers/subscribe.js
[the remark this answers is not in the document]
mcollina: I think having some standalone tests would be better.

lib/handlers/subscribe.js
```diff
@@ -78,14 +96,24 @@ function _dedupe (subs) {
...
+      })
+    })))
+    if (restore) {
+      done()
+    } else {
+      completeSubscribe.call(state)
+    }
+  } catch (err) {
+    done(err)
+  }
```
mcollina, line 116: This has the same problem with promises identified elsewhere

mcollina, line 116: This has the same problem with promises identified elsewhere

lib/handlers/subscribe.js
```diff
@@ -78,14 +96,24 @@ function _dedupe (subs) {
...
+  const state = new SubscribeState(client, packet, restore, done)
+  try {
+    await Promise.all(packet.subscriptions.map(sub => new Promise((resolve, reject) => {
+      doSubscribe.call(state, sub, (err, result) => {
+        if (err) reject(err)
+        else resolve(result)
+      })
+    })))
+    if (restore) {
+      done()
```
mcollina, line 110: Note that if this throws, it would call done twice

lib/handlers/subscribe.js
```diff
@@ -159,12 +187,15 @@ function addSubs (sub, done) {
...
 
+  /* c8 ignore start */
   if (client.closed || client.broker.closed) {
     // a hack, sometimes client.close() or broker.close() happened
     // before authenticate() comes back
     // we don't continue subscription here
+    // since it is a rare race condition we ignore it in coverage testing
     return
   }
+  /* c8 ignore stop */
```
mcollina, line 198: was this tested before?

lib/handlers/unsubscribe.js
```diff
@@ -40,13 +40,19 @@ function handleUnsubscribe (client, packet, done) {
...
+    await Promise.all(packet.unsubscriptions.map(sub => new Promise((resolve, reject) => {
+      doUnsubscribe.call(state, sub, (err, result) => {
+        if (err) reject(err)
+        else resolve(result)
+      })
+    })))
+    completeUnsubscribe.call(state)
+  } catch (err) {
+    completeUnsubscribe.call(state, err)
+  }
```
mcollina, line 55: Same problem

lib/utils.js
```diff
@@ -37,13 +37,31 @@ export function through (transform) {
...
+  return {
+    reschedule (newMs) {
+      timeout.refresh(newMs)
+    },
+    clear () {
+      clearTimeout(timeout)
     }
-  })
+  }
+}
```
mcollina, line 50: I would just remove the utility if possible.

lib/utils.js
```diff
@@ -37,13 +37,31 @@ export function through (transform) {
...
+    chunks.push(fn(chunk))
+    if (chunks.length === batchSize) {
+      yield (chunks)
+      chunks = []
+    }
+  }
+  // if chunks is half full when the iterator ends
+  if (chunks.length > 0) {
+    yield chunks
+  }
```
mcollina, line 64: Can you benchmark the simple implementation against https://github.com/mcollina/hwp/blob/main/index.js? I have a feeling this might problematic (or maybe V8 just got way faster over the years).

lib/utils.js
```diff
@@ -37,13 +37,31 @@ export function through (transform) {
...
+      clearTimeout(timeout)
     }
-  })
+  }
+}
+
+export async function * batch (readable, fn, batchSize) {
+  let chunks = []
+  for await (const chunk of readable) {
+    chunks.push(fn(chunk))
```
mcollina, line 55: I think there might be an issue with error handling here.

aedes.js
```diff
@@ -265,11 +254,14 @@ export class Aedes extends EventEmitter {
     this.closed = true
     clearInterval(this._heartbeatInterval)
     clearInterval(this._clearWillInterval)
-    this._parallel(this, closeClient, Object.keys(this.clients), doneClose)
-    function doneClose () {
+    const promises = []
+    for (const clientId of Object.keys(this.clients)) {
+      promises.push(closeClient(this.clients[clientId]))
+    }
+    Promise.all(promises).then(() => {
```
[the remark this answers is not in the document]
seriousme: replaced .then() by .finally()

lib/handlers/subscribe.js
seriousme: Not that I know, the coverage testing shows that this part is never reached and given the comments it looked like a rare race condition to me.

seriousme:
FYI: with all the changes the github review comments are a bit hard to track.
I think everything is in, if I missed anything please let me know.
I will try to benchmark `batch()` against  https://github.com/mcollina/hwp/blob/main/index.js in the coming days.

lib/utils.js
```diff
@@ -37,47 +37,38 @@ export function through (transform) {
...
-  return new Writable({
-    objectMode: true,
-    writev: function (chunks, cb) {
-      fn(chunks.map(chunk => chunk.chunk), cb)
+export async function * batch (readable, fn, batchSize) {
+  let chunks = []
+  for await (const chunk of readable) {
+    chunks.push(fn(chunk))
+    if (chunks.length === batchSize) {
+      yield (chunks)
```
robertsLando, line 64:
```suggestion
      yield chunks
```

robertsLando (changes requested):

robertsLando (approved):
Would like to wait for @mcollina feedback on this. From my POV this is good to merge 👍🏼

robertsLando:
@mcollina ping

seriousme:
any news? npm audit now warns about outdated version of uuid used by hyperid

robertsLando:
@seriousme I'm doing a review now, if that's ok I will merge it

robertsLando (commented):
## Deep code review

**Verdict:** Needs work — two real correctness bugs in the new `Promise.all`-based error paths (both flagged previously by @mcollina, both reproducible locally).

**Top 3 risks:**
1. `unsubscribe.js` — caller **hangs** on `Promise.all` rejection: `completeUnsubscribe(err)` emits `'error'` and returns *without* invoking `done`. → see inline thread reply.
2. `subscribe.js` — `done` can be called **twice** if `completeSubscribe` throws after `write()` has queued `setImmediate(done, …)`. → see inline thread reply.
3. `handlers/index.js:64` — `client._keepaliveTimer.refresh(client._keepaliveInterval)` — Node's `Timeout.refresh()` is nullary; the interval arg is silently dropped. → see inline finding.

**Strengths:**
- Direction is sound: replacing micro-deps with platform APIs (`randomUUID`, `finished`, `Timeout.refresh`) reduces supply-chain surface; benchmark shows ~+15–18% msg/s at QoS 0.
- `batch()` async generator gives bounded concurrency (default 16) and is genuinely more readable than the prior fastfall/fastparallel chain.
- `runFall` / `batch` get standalone tests in `test/utils.js`.
- `randomUUID` is **more** secure than `hyperid` (cryptographic vs. predictable); perf delta is irrelevant at once-per-connection cost.
- ObjectPool removal validated by benchmark — @mcollina's hypothesis confirmed.

**Mcollina's review comments — verification:**

| Comment | Status |
|---|---|
| `aedes.js:141` "bazillion of promises" | Addressed — `batch()` (bounded concurrency, default 16). ✅ resolving |
| `lib/utils.js:70` "use `timer.refresh`" / "remove utility" | Addressed — retimer utility removed entirely. ✅ resolving |
| `lib/utils.js:59` "ObjectPool worse than reallocating" | Addressed — removed. ✅ already resolved |
| `aedes.js:57` "verify ObjectPool needed at all" | Addressed — removed; benchmark confirmed. ✅ resolving |
| `aedes.js:138` "process.nextTick(done)" | Addressed — async/await removed the callback path. ✅ resolving |
| `aedes.js:261` "callback not invoked if one fails" | Addressed — `Promise.all(...).finally(...)`. ✅ resolving |
| `subscribe.js:22` "Please test this" | Addressed — `runFall` standalone tests in `test/utils.js`. ✅ resolving |
| `lib/utils.js:84` "benchmark vs hwp" | Addressed — benchmark posted in-thread. ✅ resolving |
| `lib/utils.js:75` "error handling here" | Addressed — `batch()` error-path test added. ✅ resolving |
| **`subscribe.js:86`** "throws → calls done twice" | ❌ **NOT FIXED** — reproducible; see thread reply. |
| **`subscribe.js:92` / `unsubscribe.js:50`** "same problem with promises" | ❌ **NOT FIXED** — unsubscribe path never calls `done` on rejection; see thread replies. |
| `subscribe.js:179` "was this tested" | ❌ NOT TESTED — explicitly `c8 ignore`d; comment to follow up. |
| Copilot +1s (`connect.js:84`, `client.js:10`) | Addressed. |

This PR is **net positive** and should land — the supply-chain reduction is the right move and the perf result is a nice bonus — but the two error-path bugs need fixing first. Both are mcollina's prior concerns and both are reproducible in ~20 lines.

lib/handlers/index.js
```diff
@@ -61,7 +61,7 @@ function handle (client, packet, done) {
   }
 
   if (client._keepaliveInterval > 0) {
-    client._keepaliveTimer.reschedule(client._keepaliveInterval)
+    client._keepaliveTimer.refresh(client._keepaliveInterval)
```
robertsLando: **[Major] · Correctness**

`Timeout.refresh()` takes **no arguments** — the `client._keepaliveInterval` arg is silently dropped. This works today only because the `setTimeout(..., _keepaliveInterval)` at `connect.js:157` set the duration once for the lifetime of the connection.

**Why:** If anyone ever needs the keepalive interval to change mid-session (e.g. to apply a renegotiated value), this call will appear to do something but won't. Latent footgun.

**Fix:** Drop the arg (`client._keepaliveTimer.refresh()`) to match Node's API; if a future use case needs a new interval, recreate the timer instead.

aedes.js
```diff
@@ -57,9 +54,7 @@ export class Aedes extends EventEmitter {
       return new Client(that, conn, req)
     }
 
-    this._parallel = parallel()
-    this._series = series()
-    this._enqueuers = reusify(DoEnqueues)
+    this._series = runSeries
```
robertsLando: **[Nit] · Readability**

`this._series = runSeries` is then used at `aedes.js:203` and `connect.js:89` via `broker._series(...)`. Since `runSeries` is a pure function with no closure over `this`, the indirection adds a hop without buying anything.

**Why:** Two slightly-different ways to call the same helper (direct import vs. broker property) makes the code base inconsistent.

**Fix:** Either call `runSeries(...)` directly at both call sites, or, if the indirection is intentional for monkey-patching/testing, drop a one-line comment saying so.

lib/utils.js
```diff
@@ -37,13 +39,49 @@ export function through (transform) {
...
+  done = (done || noop).bind(state)
+  let i = 0
+  function next (err) {
+    if (err || i === actions.length) return done(err)
+    actions[i++].call(state, packet, next)
+  }
+  next()
+}
+
+export async function * batch (readable, fn, batchSize) {
```
robertsLando, line 114: **[Minor] · Readability**

`batch()` eagerly fires `fn(chunk)` and accumulates the returned promises before yielding. That's the intended bounded-concurrency shape, but the contract is implicit.

**Why:** Callers need to `await Promise.all(batch)` to actually apply backpressure between batches. The current usage in `_clearWills` does this correctly, but the next caller might not, and the helper offers no guidance.

**Fix:** Two-line JSDoc: `fn(chunk)` runs eagerly; caller must `await Promise.all(batch)` to wait for the batch before the next one starts.

lib/handlers/unsubscribe.js
```diff
@@ -40,30 +40,36 @@ function handleUnsubscribe (client, packet, done) {
...
-  }
+async function doUnsubscribe (state, sub) {
+  return new Promise((resolve, reject) => {
+    const client = state.client
+    const broker = client.broker
+    const s = client.subscriptions[sub]
+
+    if (s) {
+      const func = s.func
+      delete client.subscriptions[sub]
```
robertsLando: **[Minor] · Correctness**

`delete client.subscriptions[sub]` runs **before** `broker.unsubscribe(...)`. If the underlying `mq.removeListener` errors (or the process crashes between the two), the client believes the topic was removed but mqemitter still holds the listener — so packets keep being delivered.

**Why:** Inconsistent client/broker state on the error path; pre-existing-ish, but the explicit `Promise` wrapper here is a good moment to swap the order.

**Fix:** Delete `client.subscriptions[sub]` **inside** the success branch of the `broker.unsubscribe` callback.

test/utils.js
```diff
@@ -0,0 +1,123 @@
...
+    }
+    t.assert.equal(numBatch, numBatches, 'all batches')
+    t.assert.equal(results.length, numItems, 'all items')
+    for (let i = 0; i < numItems; i++) {
+      t.assert.equal(results[i], i * 2)
+    }
+  })
+}
+
+test('test batch function trowing error', async (t) => {
```
robertsLando: **[Nit] · Typo**

"trowing" → "throwing".

lib/handlers/subscribe.js
robertsLando: @seriousme — the concern is **not** the `Promise.all([ok, fail])` resolution shape (you're right that the catch fires once). The issue is what happens **after** `await Promise.all(...)` resolves and `completeSubscribe.call(state)` runs.

`write()` in `lib/write.js` always uses `setImmediate(done, error, client)` — so `done` is scheduled but not yet invoked. If anything in `completeSubscribe` throws synchronously *after* that `write()` call (a listener installed on the `'subscribe'` event throws, `broker.publish(...)` throws inside an mq listener, `persistence.createRetainedStreamCombi` throws, etc.), the error escapes the try block:

```js
try {
  await Promise.all(...)
  completeSubscribe.call(state)   // queues done via write(), then throws
} catch (err) {
  done(err)                       // catch calls done
}
// next tick: setImmediate fires → done called AGAIN
```

Minimal repro:
```js
const state = { finish: function(err){ console.log('DONE', err?.message ?? 'OK') } }
function write (cb) { setImmediate(cb) }
function completeSubscribe () {
  write(state.finish)
  throw new Error('listener threw')
}
async function doSub () { return Promise.resolve() }
async function handleSubscribe () {
  try {
    await Promise.all([doSub()])
    completeSubscribe.call(state)
  } catch (err) {
    state.finish(err)
  }
}
handleSubscribe()
// → DONE listener threw
// → DONE OK     ← second invocation
```

Fix direction: move the post-`write` work outside the try block, or wrap `done` in an idempotent guard before passing it through `state.finish`.

lib/handlers/subscribe.js
robertsLando: Confirming this is still present. Same underlying issue as the double-`done` thread above — anything that throws synchronously inside `completeSubscribe` after `write()` has scheduled `done` will end up invoking `done` twice (via the `catch` here at L92 plus the queued `setImmediate`). Repro in the L86 thread reply.

lib/handlers/subscribe.js
robertsLando: Marking this as a follow-up rather than a blocker, but worth a thought: the current `/* c8 ignore */` block is justified with "rare race condition" — that's a behavioral claim that should age well, and currently the comment doesn't show the race. Two options:

1. If `client.closed || broker.closed` *can* be reached here, drop a one-liner pointing at the racing call site (e.g. *"set in `aedes.close()` while `authorizeSubscribe` is in flight; see test X"*), and ideally write the regression test.
2. If it's structurally unreachable (because `authorizeSubscribe` and `close()` are serialized somewhere upstream), then the dead branch should be removed, not ignored.

Not blocking the PR, but "rare race we ignore in coverage" is the kind of comment that rots — better to commit either way.

lib/handlers/unsubscribe.js
robertsLando: Confirming this one is still a real bug — and arguably the worst of the three since the caller **hangs forever** instead of being called twice.

In `actualUnsubscribe`:
```js
try {
  await Promise.all(packet.unsubscriptions.map(sub => doUnsubscribe(state, sub)))
  completeUnsubscribe.call(state)
} catch (err) {
  completeUnsubscribe.call(state, err)   // ← passes err
}
```

And in `completeUnsubscribe`:
```js
function completeUnsubscribe (err) {
  const client = this.client
  if (err) {
    client.emit('error', err)
    return   // ← never calls done!
  }
  ...
}
```

Result: on `Promise.all` rejection (e.g. `mq.removeListener` callback gets an err, or `broker.unsubscribe` throws), only `client.emit('error', err)` runs and `state.finish` is **never invoked**. Upstream, `handle()` in `handlers/index.js` never sees completion → `client._nextBatch` is never called → `_parsingBatch` is never decremented → that client's packet processing stalls.

Minimal repro:
```js
const state = { client: { emit: (k,e)=>console.log('emit', k, e.message), broker: {} }, packet: { unsubscriptions: ['a'] }, finish: (err) => console.log('DONE', err?.message ?? 'OK') }
async function doUnsubscribe () { return new Promise((_, reject) => setImmediate(()=>reject(new Error('mq error')))) }
function completeUnsubscribe (err) {
  const client = state.client
  if (err) { client.emit('error', err); return }
  state.finish()
}
async function actualUnsubscribe () {
  try {
    await Promise.all([doUnsubscribe()])
    completeUnsubscribe.call(state)
  } catch (err) { completeUnsubscribe.call(state, err) }
}
actualUnsubscribe()
// → emit error mq error
// → (no DONE — caller hangs)
```

Fix direction: invoke `done(err)` (i.e. `this.finish(err)`) in the catch path before emitting on the client — or, cleaner, have `completeUnsubscribe(err)` call `this.finish(err)` before the early return.

seriousme:
@robertsLando 
Ok, I'm not very fond of the LLM audits both from a quality perspective and from a community perspective, but I think everything is in now.

robertsLando (approved):

robertsLando:
@seriousme I always do a review myself and check both ai review, sometimes I pick something it miss sometimes the opposite, it's just a plus :)

lib/handlers/unsubscribe.js
robertsLando: @seriousme ?

lib/handlers/subscribe.js
robertsLando: cc @seriousme

lib/handlers/subscribe.js
robertsLando: cc @seriousme

robertsLando (changes requested):
@seriousme Please check the 3 missing open review threads

seriousme:
@robertsLando : the problem I have with Copilot is that its hard to tell which comments are yours and which are the LLM's 
E.g. when I add "Fixed" and your reply is "CC @seriousme" I don't get the response and I start to wonder:  is that really you or an LLM hallucinating?

robertsLando:
@seriousme yep understand that. BTW did you saw the 3 open threads I commented on?

seriousme:
Yes, did you see my replies?

gnought:
Thanks for making efforts.
This PR is so big and quite aggressive to minimize third-party dependencies, in addition requesting to merge to main branch.
It is a bit risky to have a big momentum in one go. Could we break the PR into smaller batches to remove dependencies one by one, so we could make sure Aedes still function in an expected way?

robertsLando:
@seriousme nope I don't see them

seriousme:
@robertsLando : strange must be a github glitch as I see them when opening a new browser window and looking at the PR. Anyway the subscribe and unsubscribe ones I fixed, you then added "cc @seriousme " after my "fixed" to which I replied "???"  as I did not understand what you ment with the "cc"  The c8 exclusion I left in as there is no test that covers this line and think I can only refactor it out if I change much more and the aim was to change as little as possible to keep the impact of the changes as small as possible. Should I mark them as resolved myself? Or do you want to check and mark them as resolved?

seriousme:
Ok, while checking if I could generate a test to solve the "c8 ignore" (which worked) I noticed that the fixes you requested add new lines not covered by testing , especially the double done scenarios. And somehow also a line in client.js.  I will see if I can get these covered as well. If not I will add "c8 ignore again"

seriousme:
Ok, I added extra tests to get coverage back to 100% again.

To do that I also had to modernize the socket destruction code which relied on the internals of nodejs
_writableState.buffer was apparently removed in Node 14.

As far as I know everything is in now. If I still missed anything please let me know.

Kind regards,
Hans

lib/client.js
```diff
@@ -170,9 +170,9 @@ class Client {
     this.conn.removeAllListeners('error')
     this.conn.on('error', noop)
     // hack to clean up the write callbacks in case of error
-    const state = this.conn._writableState
-    const list = typeof state.getBuffer === 'function' ? state.getBuffer() : state.buffer
-    list.forEach(drainRequest)
```
robertsLando, line 176: I know this was an old hacky hack to fix an annoying but, wondering if it's still legit or not, no clue if we ever have a test that covered this

lib/handlers/subscribe.js
```diff
@@ -185,13 +194,18 @@ function isStartsWithWildcard (topic) {
...
+  return function (err) {
+    /* c8 ignore next */
+    if (called) return
+    called = true
+    fn(err)
   }
+}
 
+function completeSubscribe () {
+  const done = once(this.finish)
```
robertsLando: The wrap only covers calls that go through this local `done`. The catch in `handleSubscribe` (L91) still has the outer `done` in scope and calls it directly, so the original double-call path is still reachable: `write(client, …, done)` queues `setImmediate(done, …)`, then anything throwing synchronously below — a listener on `'subscribe'`, `broker.publish(…)` reaching a throwing emitter, the `stream.pipe` setup — escapes back to the catch, which fires `done(err)`; then the queued tick fires `done(null, client)`. The flag in `once()` is on the wrapped reference; the catch path never touches that closure.

If you wrap one level up — `done = once(done)` at the top of `handleSubscribe` and drop the local wrap here — both paths share the same gate and the bug is closed.

Repro is ~15 lines if it helps:
```js
function once (fn) { let c = false; return e => { if (c) return; c = true; fn(e) } }
const state = { finish: e => console.log('DONE', e?.message ?? 'OK') }
function completeSub () {
  const done = once(state.finish)
  setImmediate(done)               // mimics write(...,done)
  throw new Error('listener threw') // anything sync after write
}
async function handleSub (done) {
  try { await Promise.resolve(); completeSub() }
  catch (err) { done(err) }        // bypasses once()
}
handleSub(state.finish)
// → DONE listener threw
// → DONE OK
```

lib/handlers/unsubscribe.js
```diff
@@ -40,42 +40,50 @@ function handleUnsubscribe (client, packet, done) {
...
-    doUnsubscribe,
-    packet.unsubscriptions,
-    completeUnsubscribe)
+async function actualUnsubscribe (client, packet, done) {
+  const state = new UnsubscribeState(client, packet, done)
+  try {
+    await Promise.all(packet.unsubscriptions.map(sub => doUnsubscribe(state, sub)))
+    completeUnsubscribe.call(state)
+  } catch (err) {
+    completeUnsubscribe.call(state, err)
```
robertsLando, line 49: The new `return done(err)` in `completeUnsubscribe` closes the "caller hangs" path on Promise.all rejection — good.

But the symmetric case now opens up the same shape as the subscribe thread: on the success path, `write(client, UnSubAck, done)` (L89) queues `setImmediate(done)`, then `client.broker.emit('unsubscribe', …)` (L96) can throw if a listener throws — escapes `completeUnsubscribe`, lands here in the catch → `completeUnsubscribe.call(state, err)` → `done(err)` at L83. The queued tick then fires `done(null, client)` a second time.

Same fix shape: wrap `done` once at the top of `actualUnsubscribe` (`done = once(done)`) so both branches share one gate. The `once` helper added in `subscribe.js` could move to `utils.js` and be reused here.

robertsLando (commented):
Two follow-ups on the error-path fixes from `e366386`.

seriousme:
@robertsLando : as far as I know I have adressed all open issues, hence I requested rereview 3 weeks ago. Is there anything I still need to do?

robertsLando:
@seriousme I'm so sorry for the wait time on this, I wanted to wait for a last @mcollina review on this but since he is very busy I will give it a last check now, sorry again 🙏🏼

seriousme:
@robertsLando 
No worries, just want to make sure that I don't need to refactor this PR again because of the new MQTTv5 PR 
(and github keeps nagging me every week about aedes using a vulnerable version of UUID 😉 )

robertsLando:
@seriousme I pushed a small commit (ae61fd6) directly to the branch — it adds standalone unit tests for `once()` and `runSeries()` in `test/utils.js`, plus a one-word clarification to the `c8 ignore` comment on the `once()` guard. This closes the test-coverage gap @mcollina asked about (double-callback prevention + series error-stop/sequencing semantics). Tests pass (15/15) and lint is clean. Feel free to amend/squash as you see fit.

On the review status: I went over all the threads and from my side the correctness items are addressed. The one remaining point is @mcollina's note about `Promise.all` in subscribe/unsubscribe — worth confirming whether that's still a blocker for you, given the publish hot path still uses callback-based `runSeries`.

robertsLando:
Pushed a570cc0: replaced `Promise.all(subscriptions.map(...))` in the subscribe/unsubscribe handlers with a local callback-based `runParallel()` helper in `utils.js` (sibling to `runFall`/`runSeries`), addressing @mcollina's note about per-packet promise allocation.

- Fans the per-topic callbacks out and invokes `done` once — no `new Promise` per topic, so allocation stays flat on the control path (matches the original `fastparallel` behaviour this PR set out to replace).
- Behaviour-preserving vs the previous `Promise.all`: actions still run concurrently, `granted` ordering is kept via the synchronous `subState.push`, error semantics (first-error wins) are unchanged, and the `once()`-wrapped `done` still guards double-invocation.
- Added standalone `runParallel` unit tests (success, first-error + straggler-ignored, `this`-binding, empty array). Full subscribe/unsubscribe suites pass.

Note: `runParallel` matches the PR's existing `Promise.all` error timing (report on first error) rather than fastparallel's wait-for-all — happy to switch to wait-for-all if you'd prefer the exact original semantics.

seriousme:
Thx!

PR #1056, opened 2025-10-16
fix: ensure broker always respond with PUBCOMP to a PUBREL in QoS 2

### LINKED ISSUE #634: [bug] `pubrel` delivery re-try causes `no such packet` error
**System Information**
- Aedes: 0.42.5
- NodeJS: v14.15.4
- OS: MacOS 10.15.7
- Arch: x86_64

**Describe the bug**
When receiving `pubrel` packet for a message that's been already fully processed from the broker's point of view, the broker logs `no such packet` error and disconnects. The client. If the client didn't receive the initial `pubcomp` and wants to resend the `pubrel` upon reconnection, this will lead to a re-connection loop, where the client would connect, send `pubrel`, get disconnected, re-connect, send `pubrel`, get disconnected and so on.

**To Reproduce**
Run the `publish QoS 2, pubrel delivery retry` test I've added [here](https://github.com/tomekt-bolt/aedes/commit/18d94a86e8beebe160447581ad83b1ac8333fa35#diff-71ded4935931e595578532a61938d1896a18910163647103e465d2a830680325R97).

**Expected behavior**
The broker should neither emit any error nor disconnect the client when receiving `pubrel` packet for already `pubcomp`ed message. It should send the `pubcomp` once more instead.

**Additional context**
According to the [MQTT spec](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718103) the client
> MUST re-send any unacknowledged PUBLISH Packets (where QoS > 0) and PUBREL Packets using their original Packet Identifiers

so conforming to this should not cause any errors.

robertsLando:
Fixes #634

### DIFF
--- a/lib/handlers/publish.js
+++ b/lib/handlers/publish.js
@@ -47,12 +47,16 @@ function enqueuePublish (packet, done) {
           // Duplicate packet: just send PUBREC, don't publish again
           write(client, new PubRec(packet), done)
         }, () => {
-          // New packet (not found in store): publish it and store it
-          client.broker.publish(packet, client, (err) => {
-            if (err) { return done(err) }
-            client.broker.persistence.incomingStorePacket(client, packet)
-              .then(() => write(client, new PubRec(packet), done), done)
-          })
+          // New packet (not found in store): store first, then publish
+          // This ensures if storage fails, message hasn't been delivered yet
+          // preventing duplicate delivery on retransmission
+          client.broker.persistence.incomingStorePacket(client, packet)
+            .then(() => {
+              client.broker.publish(packet, client, (err) => {
+                if (err) { return done(err) }
+                write(client, new PubRec(packet), done)
+              })
+            }, done)
         })
       break
     case 1:


seriousme:
Just checked: the tests fail on QoS2 handling,which makes sense because the QoS2  message has not been published yet:

https://github.com/moscajs/aedes/blob/2c6882ef60f1b41fd5d5d2d36168eaa100afe6fd/lib/handlers/publish.js#L42-L46

What needs to happen in the publish handler (per[MQTT-4.3.3-2]) is:
a) check if we already have packetID  by using client.broker.persistence.incomingGetPacket()
b) if not found:
 - client.broker.publish()  // we have taken ownership for the packet so we publish it
 - if publish was successfull : client.broker.persistence.incomingStorePacket()
 c) always send PUBREC packet for this packet , even if we already have it in persistence

This way the same publish packet will never be published twice

Then in the PubRel handler we:
a) check if we have packetID  by using client.broker.persistence.incomingGetPacket()
b) if  found, we delete this packetID by using persistence.incomingDelPacket()
c) always send a PUBCOMP packet for this packet, even if we did not have it in persistence

This releases the packetID so that it can be reused for a new PUBLISH packet.

Hope this helps!

Hans
ps. further optimization would be not to store the whole packet in incomingStore, but only the packetID. since we only need to retain the packetID.

test/topics.js
```diff
@@ -173,91 +173,123 @@ async function subscribeMultipleTopics (t, publisher, subscriber, qos, subscript
...
-    { topic: 'hello/world', qos: 1 },
-    { topic: 'hello/#', qos: 1 }]
-  await subscribeMultipleTopics(t, publisher, subscriber, 1, sub)
   const packet = await nextPacket(subscriber)
   delete packet.messageId
   t.assert.deepEqual(structuredClone(packet), expected, 'packet must match')
 })
 
 // [MQTT-3.3.5-1]
+// FIXME: The broker currently delivers at first subscription's QoS (0), not max QoS (2)
```
robertsLando: Let's keep this for another issue

test/not-blocking.js
```diff
@@ -106,8 +103,12 @@ for (const [title, brokerOpts, subscription] of
         console.log('sent / received', sent, received)
       }
       received++
+      if (received === total) {
+        // Close subscriber when all messages received
+        setImmediate(() => subscriber.end())
+      }
     })
-    subscriber.subscribeAsync(subscription)
+    await subscriber.subscribeAsync(subscription)
```
robertsLando: this actually fixes a super weird issue that causes this tests to not work on my instance both with node 20-22 (even on main)

Issue #821, opened 2023-05-03
[feat] Tracking issue MQTTv5


robertsLando:
The features that doesn't require changes on persistences are: 

- [ ]  [Reason code](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901031) on all ACKs
- [ ]  Reason string on all ACKs
- [ ]  Message expiry: the lifetime of the Will Message in seconds
- [ ]  Server disconnect
- [ ]  Payload format and content type (this should be already handled by mqtt-packet we just have to forward it)
- [ ]  [Request / Response](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901252)
- [ ]  Subscription id (persistence `addSubscription` stores the entire sub packet so we don't have to make edits on that side)
- [ ]  Flow control
- [ ]  User properties (this should be already handled by mqtt-packet we just have to forward it)
- [ ]  Maximum Packet Size
- [ ]  Optional Server feature availability
- [ ]  Enhanced authentication
- [ ]  [Subscription options](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901169): I think this should already be implemented, expecially the noLocal flag but we should check this
- [ ]  Will delay: checking the `putWill` we store the entire packet in persistence so no change is needed on that part
- [ ]  Server Keep Alive
- [ ]  Assigned ClientID
- [ ]  Server reference

The ones that require changes on persistences:

- [ ] Session expiry: we actually don't have storage support for this
- [ ] Shared subscriptions: I drafted an implementation in #449 but i never ended it
- [ ] Topic alias

robertsLando:
✅ **#1088 merged into `mqttv5`** on 2026-06-30 (merge commit `e7d0ce9`).

- Companion releases are out, so `package.json` now uses published ranges (`aedes-packet@^4.0.1`, `aedes-persistence@^10.3.1`); the `overrides` shim and the git-dep release block are gone.
- Because #1088 targeted `mqttv5` (not the default branch), the per-feature issues it referenced were **not** auto-closed — the checklist at the top continues to track them until `mqttv5` lands for release under #194.

The development-status section above has been updated to reflect the merge.

BenjaminDobler:
## Proposed sequence for the remaining **no-persistence** items

[...]

PR #1094, opened 2026-06-30
ci: MQTT 5.0 + 3.1.1 compatibility workflow (Paho interop)

robertsLando:
## What

Adds a **MQTT compatibility** CI workflow that measures aedes's *actual* MQTT 5.0
and 3.1.1 protocol compatibility by running the vendor-neutral
[Eclipse Paho interoperability suite](https://github.com/eclipse-paho/paho.mqtt.testing)
(the same suite Mosquitto, EMQX and mochi-mqtt self-certify against) and reports a
pass percentage as a sticky PR comment + job summary.

Unlike `test/mqtt5.js` (which drives aedes with the lenient `mqtt.js` client against
our own reading of the spec), Paho uses a strict, independent client and its own
packet de/serialiser — so it surfaces real interop gaps our own tests miss.

Part of the MQTT 5.0 work (#821).

## How it works

```
aedes (this branch, tools/mqtt-compat/broker.js, :1883)
   ▲ TCP
run_compat.py ──> Paho client_test5.py (v5, 27 cases) + client_test.py (v3.1.1, 10)
   ├─ report.md  → sticky PR comment (<!-- MQTT_COMPAT_COMMENT -->) + job summary
   └─ result.json → uploaded artifact
```

- **`tools/mqtt-compat/broker.js`** — boots aedes on a TCP port with the v5
  capabilities the suite exercises enabled (inbound topic aliases) plus an
  `authorizeSubscribe` rule that the `test_subscribe_failure` case requires
  (matching the Paho reference broker's config).
- **`tools/mqtt-compat/run_compat.py`** — runs each Paho test in its **own
  subprocess with a hard timeout** (the Paho clients block with no internal
  timeout, so an unsupported feature would otherwise hang the whole run). Pure
  stdlib. Emits a Markdown report + JSON.
- **`.github/workflows/mqtt-compat.yml`** — runs on `pull_request` and
  `workflow_dispatch`; clones the suite at a pinned SHA, runs it, posts the report.

## Results on this branch

| Protocol | Compatibility |
|----------|---------------|
| MQTT 5.0 | **62%** (16/26 evaluated) |
| MQTT 3.1.1 | **90%** (9/10) |

The percentage is **% of the Paho functional suite**, not "% v5-compliant" — see
the coverage note below. Failures fall into three buckets:

**Expected gaps** (intentional / deferred-by-design per #821, annotated, still
counted as fails):
- `shared_subscriptions` — advertised as unavailable (`sharedSubscriptionAvailable=false`; #841).
- `flow_control1` — `receiveMaximum` advertised but not enforced outbound (#829).
- `server_topic_alias` — broker-assigned (outbound) topic aliases not implemented;
  inbound aliases work (#840, spec-optional/MAY).
- `subscribe_identifiers` — a delivery matching multiple overlapping subscriptions
  echoes only one Subscription Identifier (#828, [MQTT-3.3.4-4]).

**Real interop gaps surfaced** (the Paho reference broker passes all of these under
the identical harness — genuine differences from a spec-compliant broker):
- `will_message` / `will_delay` — aedes delivers the will but **leaks the
  `WillDelayInterval` will-property into the delivered PUBLISH**, which strict
  clients reject (`Property WillDelayInterval does not apply to packet type Publish`).
  #821 marked #835 as conformant; this contradiction is filed as a bug: **#1098**.
- `maximum_packet_size` — the client's declared Maximum Packet Size isn't honoured
  on outbound (broker delivers an oversized message instead of dropping it; #831
  documents outbound enforcement as advisory-only).
- `offline_message_queueing`, `publication_expiry` — v5 session/expiry edge cases.
- **v3.1.1 `zero_length_clientid`** — empty ClientId + CleanSession=0 is accepted
  instead of rejected with code 0x02. Filed as a bug: **#1095**.
- Note: setting `keepaliveLimit` makes `server_keep_alive` pass but currently breaks
  **every** MQTT 3.1.1 CONNECT (reason code 6) — so it's left disabled. Worth fixing.

**Harness-limited** (excluded from the denominator, not a broker issue):
- `flow_control2` — not evaluable under per-test isolation; the Paho reference broker
  also hangs it here. The capability it probes is still covered by `flow_control1`.

### Self-flagging stale gaps (v5 is WIP)

When a test currently in `EXPECTED_GAPS` starts **passing** (someone implemented the
feature), the report highlights it with a 🎉 banner, the row is marked, and the
workflow emits a CI `::warning::` — prompting a maintainer to drop the now-stale
entry from `EXPECTED_GAPS`.

### Coverage scope (important)

The Paho suite is a third-party **happy-path functional** test. It's a strong
cross-implementation signal for user-facing features, but it touches only 5 reason
codes and omits aedes's error/negative surface (most of the 19 reason codes in
`lib/constants.js`) and whole spec chapters (AUTH/§4.12, Server Reference, CONNACK
capability flags, ReasonString). Those are covered by `test/mqtt5.js`, so the two
are **complementary**: Paho = functional cross-impl gauge, `test/mqtt5.js` =
authoritative error-semantics. The conformance area left uncovered by *both* (AUTH
robustness + ReasonString / capability-flag round-trips) is tracked in **#1096**.

### Harness validation

Validated against the Paho project's own fully-compliant reference broker
(`startbroker.py`): **26/26** evaluated v5 tests pass under this harness. That
calibration is what makes the aedes failures above trustworthy rather than artifacts.

## Notes

- Because GitHub evaluates `pull_request` workflows from the **base** branch, this
  workflow won't auto-run on *this* PR until it's on `mqttv5` — use **Run workflow**
  (`workflow_dispatch`), or it runs on the next PR after merge.
- Non-blocking by design: it reports the score and warns; it doesn't fail the build
  on known gaps.

Follow-up issues filed: **#1095** (v3.1.1 zero-length-clientId bug), **#1096** (v5
AUTH/ReasonString test-coverage gap).

### DIFF
--- a/.github/actions/sticky-pr-comment/action.yml
+++ b/.github/actions/sticky-pr-comment/action.yml
@@ -19,12 +19,17 @@ runs:
         echo "EOF" >> $GITHUB_OUTPUT
 
     - name: Post or update PR comment
-      uses: actions/github-script@v7
+      uses: actions/github-script@v9
+      env:
+        # Pass via env, not script interpolation: the file body can contain
+        # backticks or ${...} that would otherwise break the JS template literal.
+        STICKY_TAG: ${{ inputs.tag }}
+        STICKY_BODY: ${{ steps.file.outputs.body }}
       with:
         github-token: ${{ github.token }}
         script: |
-          const tag = `${{ inputs.tag }}`;
-          const body = `${tag}\n${{ steps.file.outputs.body }}`;
+          const tag = process.env.STICKY_TAG;
+          const body = `${tag}\n${process.env.STICKY_BODY}`;
           const pr = context.payload.pull_request?.number;
           if (!pr) {
             core.setFailed("No pull request found in context.");
--- a/.github/actions/sticky-pr-comment/action.yml
+++ b/.github/actions/sticky-pr-comment/action.yml
@@ -13,10 +13,20 @@ runs:
     - name: Read file and set as output
       id: file
       shell: bash
+      env:
+        # Reference the input via env, never splice ${{ inputs.file }} into the
+        # script: a caller-controlled path must not be able to inject shell.
+        INPUT_FILE: ${{ inputs.file }}
       run: |
-        echo "body<<EOF" >> $GITHUB_OUTPUT
-        cat "${{ inputs.file }}" >> $GITHUB_OUTPUT
-        echo "EOF" >> $GITHUB_OUTPUT
+        # Random, unguessable heredoc delimiter: a fixed "EOF" would let a body
+        # line that is literally "EOF" terminate the heredoc early and inject
+        # arbitrary key=value pairs into $GITHUB_OUTPUT.
+        delimiter="ghadelimiter_$(openssl rand -hex 16)"
+        {
+          echo "body<<$delimiter"
+          cat "$INPUT_FILE"
+          echo "$delimiter"
+        } >> "$GITHUB_OUTPUT"
 
     - name: Post or update PR comment
       uses: actions/github-script@v9
--- a/.github/workflows/mqtt-compat.yml
+++ b/.github/workflows/mqtt-compat.yml
@@ -72,6 +72,24 @@ jobs:
         if: always()
         run: kill "${BROKER_PID}" 2>/dev/null || true
 
+      - name: Warn on unexpected passes
+        if: always()
+        run: |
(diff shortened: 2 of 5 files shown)


robertsLando:
cc @BenjaminDobler 

The report surfaced an issue in mqtt 3.1 :) https://github.com/moscajs/aedes/pull/1094#issuecomment-4843397009

Issue #1092, opened 2026-06-29
Request Problem Information (CONNECT 0x17)


robertsLando:
Part of the MQTT 5.0 tracking issue #821.

## Feature

MQTT 5.0 **Request Problem Information** — CONNECT property `0x17` (§3.1.2.11.7), default `1`.

It gates whether the server may attach a **Reason String** and/or **User Properties** to packets **other than** PUBLISH, CONNACK, and DISCONNECT (i.e. PUBACK/PUBREC/PUBREL/PUBCOMP/SUBACK/UNSUBACK/AUTH). If the client sets it to `0`, the server **MUST NOT** send a Reason String or User Properties on those packets ([MQTT-3.1.2-29]).

## Relationship to #823 / #830

This is a gating constraint on:
- #823 — Reason String on ACKs
- #830 — User Properties

It must be honored once either of those is implemented for the affected packet types.

## Current status

Benign today: the broker emits neither `reasonString` nor `userProperties` on any outbound packet (verified against the `mqttv5` branch / PR #1088), so there is currently nothing to suppress and no violation. It becomes a real obligation as soon as #823 / #830 extend Reason String / User Properties to ACKs — at which point those emissions must be gated on this flag. Tracked here so the constraint isn't missed.

PR #1100, opened 2026-06-30
feat: MQTT 5.0 reason codes & reason strings on ACKs (#822, #823, #1092)

BenjaminDobler:
PR 1 of the post-merge #821 follow-ups, per the [agreed sequence](https://github.com/moscajs/aedes/issues/821#issuecomment-4842940944). Broker-core only — no persistence changes, cluster-safe.

## #1092 — Request Problem Information [MQTT-3.1.2-29]
Parse and store `client._requestProblemInformation` (mqtt-packet decodes the byte property as a **boolean**; absent ⇒ `true`). When `false`, the broker must not send a Reason String or User Properties on packets other than PUBLISH / CONNACK / DISCONNECT.

## #823 — Reason String on ACKs (auto-sourced from broker errors)
An unauthorized v5 publish now carries the `authorizePublish` `error.message` as the **Reason String** on its `0x87` PUBACK/PUBREC, gated by Request Problem Information. New `ackProperties(client, reasonString)` helper in `lib/utils.js` centralizes the v5 + gating check (returns `undefined` so no properties block is emitted when not applicable).

_Per the agreed design, reason strings are sourced automatically from the broker's existing errors — no new API surface._

## #822 — Granular reason codes on ACKs
- A denied SUBSCRIBE returns the specific **`0x87`** (Not authorized) for v5 clients instead of the coarse `0x80` (v3/v4 still get `0x80`).
- PUBREL / PUBCOMP carry an explicit **`0x00`** success reason code (v3/v4 ignore it; absent ≡ `0x00` on the wire, so this is explicit-only).

## Tests
- PUBACK Reason String present by default + **suppressed** when the client sets Request Problem Information = `false`.
- A denied SUBSCRIBE → `0x87` on the SUBACK.

## Docs
`authorizePublish` / `authorizeSubscribe` handler notes documenting the v5 behavior (no connection close on authz failure; error message → Reason String; SUBACK `0x87`).

369 unit tests + lint + tsd green.

### DIFF
--- a/lib/handlers/publish.js
+++ b/lib/handlers/publish.js
@@ -1,4 +1,4 @@
-import { runSeries } from '../utils.js'
+import { runSeries, ackProperties } from '../utils.js'
 import write from '../write.js'
 import { ReasonCodes } from '../constants.js'
 
@@ -125,6 +125,10 @@ function authorizePublish (packet, done) {
     if (packet.qos > 0) {
       const ack = packet.qos === 1 ? new PubAck(packet) : new PubRec(packet)
       ack.reasonCode = ReasonCodes.NOT_AUTHORIZED
+      // [#823] Reason String, sourced from the authorization error and gated by
+      // the client's Request Problem Information.
+      const properties = ackProperties(client, err.message)
+      if (properties) ack.properties = properties
       write(client, ack, () => done())
     } else {
       done()
--- a/lib/handlers/pubrec.js
+++ b/lib/handlers/pubrec.js
@@ -1,9 +1,12 @@
 import write from '../write.js'
+import { ReasonCodes } from '../constants.js'
 
 class PubRel {
   constructor (packet) {
     this.cmd = 'pubrel'
     this.messageId = packet.messageId
+    // [#822] MQTT 5.0 success reason code (0x00); ignored when serializing v3/v4.
+    this.reasonCode = ReasonCodes.SUCCESS
   }
 }
 
--- a/lib/handlers/pubrec.js
+++ b/lib/handlers/pubrec.js
@@ -11,6 +11,18 @@ class PubRel {
 }
 
 function handlePubrec (client, packet, done) {
+  // MQTT 5.0 §4.3.3: a PUBREC whose reason code is >= 0x80 terminates the QoS 2
+  // flow — the broker must NOT send a PUBREL. It releases the packet identifier
+  // (clearing the stored outgoing state) instead of continuing the handshake.
+  if (client.version === 5 && packet.reasonCode >= 0x80) {
+    if (client.clean) {
+      return done()
+    }
+    client.broker.persistence.outgoingClearMessageId(client, packet)
+      .then(() => done(), done)
+    return
+  }
+
   const pubrel = new PubRel(packet)
 
   if (client.clean) {
--- a/lib/handlers/subscribe.js
+++ b/lib/handlers/subscribe.js
@@ -161,8 +161,10 @@ function blockDollarSignTopics (func) {
 
 function storeSubscriptions (sub, done) {
   if (!sub || typeof sub !== 'object') {
-    // means failure: MQTT 3.1.1 specs > 3.9.3 Payload
-    this.granted = 128
+    // Subscription denied by authorizeSubscribe. [#822] v5 reports the specific
+    // 0x87 (Not authorized); v3/v4 SUBACK only has the coarse 0x80 failure code
+    // (MQTT 3.1.1 §3.9.3 Payload).
+    this.granted = this.client.version === 5 ? ReasonCodes.NOT_AUTHORIZED : 128
     return done(null, null)
   }
 
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -2,6 +2,19 @@ import { Transform } from 'stream'
 
 export function noop () { }
 
+// MQTT 5.0 Request Problem Information [MQTT-3.1.2-29]: build the `properties`
+// block carrying a Reason String for an outgoing ACK (PUBACK / PUBREC / PUBREL /
+// PUBCOMP / SUBACK / UNSUBACK). Returns undefined — so no properties block is
+// emitted — for non-v5 clients, when there is no reason string, or when the
+// client set Request Problem Information to 0 (it then must not receive a Reason
+// String or User Properties on packets other than PUBLISH/CONNACK/DISCONNECT).
+export function ackProperties (client, reasonString) {
+  if (client.version !== 5 || !reasonString || client._requestProblemInformation === false) {
+    return undefined
+  }
+  return { reasonString }
+}
+
 // Node clamps setTimeout delays above 2^31-1 ms (~24.8 days) to 1 ms, firing
 // almost immediately. MQTT 5.0 session-expiry / will-delay intervals are uint32
 // seconds (up to ~136 years), so a raw setTimeout would wipe long-lived sessions
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -3,11 +3,12 @@ import { Transform } from 'stream'
 export function noop () { }
 
 // MQTT 5.0 Request Problem Information [MQTT-3.1.2-29]: build the `properties`
-// block carrying a Reason String for an outgoing ACK (PUBACK / PUBREC / PUBREL /
-// PUBCOMP / SUBACK / UNSUBACK). Returns undefined — so no properties block is
-// emitted — for non-v5 clients, when there is no reason string, or when the
-// client set Request Problem Information to 0 (it then must not receive a Reason
-// String or User Properties on packets other than PUBLISH/CONNACK/DISCONNECT).
+// block carrying a Reason String for an outgoing ACK. Currently the PUBACK /
+// PUBREC error path uses it; it's the single gate to route any other ACK's
+// Reason String through as they adopt one. Returns undefined — so no properties
+// block is emitted — for non-v5 clients, when there is no reason string, or when
+// the client set Request Problem Information to false (it then must not receive a
+// Reason String or User Properties on packets other than PUBLISH/CONNACK/DISCONNECT).
 export function ackProperties (client, reasonString) {
   if (client.version !== 5 || !reasonString || client._requestProblemInformation === false) {
     return undefined
--- a/.github/workflows/mqtt-compat.yml
+++ b/.github/workflows/mqtt-compat.yml
@@ -51,6 +51,19 @@ jobs:
           git clone --filter=blob:none --no-checkout https://github.com/eclipse-paho/paho.mqtt.testing paho
           git -C paho checkout "$PAHO_REF"
 
+      # Apply targeted fixes for known-stale assertions in the pinned suite (e.g.
+      # a v5 test still asserting a 3.1.1-only reason code). Each patch documents
+      # its rationale in its header. Fails loudly if a patch no longer applies —
+      # that is the signal to re-anchor it when bumping PAHO_REF.
+      - name: Patch pinned Paho suite
+        run: |
+          shopt -s nullglob
+          for p in tools/mqtt-compat/patches/*.patch; do
+            echo "::group::Applying $(basename "$p")"
+            git -C paho apply --verbose "$GITHUB_WORKSPACE/$p"
+            echo "::endgroup::"
+          done
+
       - name: Start aedes broker
         run: |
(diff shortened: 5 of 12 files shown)


lib/handlers/pubrec.js
```diff
@@ -1,9 +1,12 @@
 import write from '../write.js'
+import { ReasonCodes } from '../constants.js'
 
 class PubRel {
   constructor (packet) {
     this.cmd = 'pubrel'
     this.messageId = packet.messageId
+    // [#822] MQTT 5.0 success reason code (0x00); ignored when serializing v3/v4.
+    this.reasonCode = ReasonCodes.SUCCESS
```
robertsLando, line 9: **[Minor] · Correctness / Spec**

`handlePubrec` unconditionally builds a PUBREL with `reasonCode = SUCCESS`, but per MQTT 5.0 §4.3.3 / §2.2.1 when the broker receives a **PUBREC whose Reason Code is ≥ 0x80**, it must **not** send PUBREL — the QoS 2 flow terminates and the Packet Identifier is released immediately.

**Why:** a subscriber that rejects a delivered QoS 2 PUBLISH (e.g. returns `0x87` in its PUBREC) will still receive a PUBREL, and the packet id isn't freed as the spec requires. This is pre-existing behavior, but this PR is the natural place to formalize it since it's now setting the PUBREL reason code explicitly.

**Fix:** in `handlePubrec`, short-circuit when `packet.reasonCode >= 0x80` (v5 clients): skip the PUBREL, release the packet id / outgoing state, and don't call the persistence update.

lib/utils.js
```diff
@@ -2,6 +2,19 @@ import { Transform } from 'stream'
 
 export function noop () { }
 
+// MQTT 5.0 Request Problem Information [MQTT-3.1.2-29]: build the `properties`
+// block carrying a Reason String for an outgoing ACK (PUBACK / PUBREC / PUBREL /
```
robertsLando: **[Nit] · Readability**

The doc-comment advertises `ackProperties` for "PUBACK / PUBREC / PUBREL / PUBCOMP / SUBACK / UNSUBACK", but it's currently only wired into PUBACK/PUBREC (`publish.js`). The SUBACK denial path sets `granted = 0x87` without going through this helper.

**Why:** a reader will grep for the helper expecting it on all six ACK types and not find it, or assume SUBACK reason strings are handled here when they aren't.

**Fix:** either narrow the comment to the ACK types actually using it today, or note it's the intended single gate and wire the remaining ACK paths through it.

lib/handlers/subscribe.js
```diff
@@ -161,8 +161,10 @@ function blockDollarSignTopics (func) {
 
 function storeSubscriptions (sub, done) {
   if (!sub || typeof sub !== 'object') {
-    // means failure: MQTT 3.1.1 specs > 3.9.3 Payload
-    this.granted = 128
+    // Subscription denied by authorizeSubscribe. [#822] v5 reports the specific
+    // 0x87 (Not authorized); v3/v4 SUBACK only has the coarse 0x80 failure code
+    // (MQTT 3.1.1 §3.9.3 Payload).
+    this.granted = this.client.version === 5 ? ReasonCodes.NOT_AUTHORIZED : 128
```
robertsLando, line 167: **[FYI] · Design / Spec**

Every `null` return from `authorizeSubscribe` (the documented "negate a subscription" path) now maps to `0x87 Not authorized` for v5 clients, even when the app nulled the sub for a non-authorization reason (e.g. an unsupported/undesired filter).

**Why:** `0x80 Unspecified error` was semantically broader; `0x87` asserts an authz decision the app may not have intended. Reasonable given the hook is named `authorizeSubscribe`, so no change strictly required — just confirm this narrowing is intended and matches the docs update.

lib/handlers/publish.js
```diff
@@ -125,6 +125,10 @@ function authorizePublish (packet, done) {
     if (packet.qos > 0) {
       const ack = packet.qos === 1 ? new PubAck(packet) : new PubRec(packet)
       ack.reasonCode = ReasonCodes.NOT_AUTHORIZED
+      // [#823] Reason String, sourced from the authorization error and gated by
+      // the client's Request Problem Information.
+      const properties = ackProperties(client, err.message)
```
robertsLando, line 130: **[FYI] · Security / Info-disclosure**

The broker's `authorizePublish` `error.message` is sent verbatim to the client as the ACK Reason String.

**Why:** if a handler puts internal detail into the error (paths, user ids, backend messages), it now leaks to the publisher. It's opt-in (the handler chooses the message) and gated by Request Problem Information, so this is by design (#823) — flagging so the docs make clear the message is client-visible and handlers should keep it generic.

robertsLando (changes requested):
## Deep review — MQTT 5.0 reason codes & reason strings on ACKs

**Verdict:** Approach is sound and spec-correct; requesting changes so the inline threads below get addressed before merge.

**Verified against OASIS MQTT 5.0 (os):**
- ✅ `requestProblemInformation` gating — mqtt-packet decodes the byte as a **boolean** (`parser.js`: `case 'byte': return this._parseByte() !== 0`), so absent⇒`undefined`, wire-0⇒`false`; the `!== false` / `=== false` gates are correct. [MQTT-3.1.2-29] honored (Reason String suppressed on PUBACK/PUBREC/SUBACK when RPI=0; PUBLISH/CONNACK/DISCONNECT unaffected).
- ✅ SUBACK `0x87` — valid for SUBACK per Table 2-6; v3/v4 correctly kept at `0x80`.
- ✅ PUBREL/PUBCOMP `0x00` and v3/v4 safety — `writeToStream.js confirmation()` writes the reason byte only for `version === 5 && length !== 2`, and v5+0x00+no-props collapses to length 2 (omitted on wire). The "ignored on v3/v4" comments are accurate.
- ✅ Per-client serialization plumbing via `client._wireVersion` → frozen `WRITE_OPTS` is correct.

**Strengths:** `ackProperties` centralizes the v5+RPI+reasonString gate (good DRY); comments cite the exact `[MQTT-…]` clauses; tests cover both present-by-default and suppressed-when-RPI=0. New tests pass in isolation (2/2).

**Threads to address:** one Minor spec gap (PUBREL on error PUBREC), one Nit (helper doc), and two FYIs (0x87 for all null subs; error-message disclosure). See inline comments.

BenjaminDobler:
Thanks for the review 🙏 — addressed in `2f6f5bb`.

**[Minor · Correctness/Spec] `pubrec.js` — no PUBREL on a rejecting PUBREC** — fixed. `handlePubrec` now short-circuits for v5 when the received PUBREC's reason code is `>= 0x80` [MQTT 5.0 §4.3.3]: it does **not** send a PUBREL, and instead releases the packet identifier (`outgoingClearMessageId`, clearing the stored outgoing state) — for clean clients it just finishes. Added a test: a raw v5 subscriber rejects the delivered QoS 2 PUBLISH with a `0x87` PUBREC and asserts no PUBREL follows.

**[Nit · Readability] `utils.js` — `ackProperties` doc-comment** — narrowed to the ACK path that actually uses it today (PUBACK/PUBREC), noting it's the intended single gate for Reason Strings as other ACK types adopt one.

**[FYI · Design/Spec] `subscribe.js` — `0x87` for every null sub** — confirming this is intended. The hook is `authorizeSubscribe`, so a `null` return (the documented "negate") maps to `0x87` (Not authorized) for v5; v3/v4 keep `0x80`. The docs note this. If you'd prefer a neutral `0x80` default (or making the code app-selectable), happy to change — but `0x87` reads as the most accurate for an authorization hook.

**[FYI · Security/Info-disclosure] `publish.js` — error message → Reason String** — by design (#823), opt-in and RPI-gated, and I've hardened the docs: the `authorizePublish` note now states the message is **client-visible** and that handlers should keep it generic (no paths/user-ids/backend detail).

370 unit tests + lint + tsd green. I've also rebased the stacked PRs on top of this fix — #1102 (`9dbb847`) and #1115 (`1c8b075`) — so they stay current.

robertsLando:
@BenjaminDobler I have merged some changes on branch mqttv5 could you please sync changes on the opened pr?

BenjaminDobler:
Done 👍 — rebased onto the latest `mqttv5` (`9938b2b`, incl. the zero-length ClientId fix #1097 and the compat workflow #1094). Conflict-free. Synced the whole stack:

- #1100 `feat/v5-ack-reason-codes` → `8a3be6b`
- #1102 `feat/v5-request-response-info` → `279160b`
- #1115 `feat/v5-server-reference` → `a4d210f`

lint + tsd + full unit green on each. The new Paho compat workflow is a great addition — happy to add v5 reason-code / server-redirect cases to it as these land if useful.

robertsLando (commented):
## Compatibility report regresses silently — harness needs updating in this PR

The code change is correct, but the **MQTT Compatibility** job is now reporting a misleading number, and the PR doesn't touch `tools/mqtt-compat/` to account for it.

### What happens

| Run | MQTT 5.0 | `test_subscribe_failure` (v5) |
|---|---|---|
| `mqtt5-compat-tests` baseline (pre-`0x87`) | **62%** (16/26) | ✅ pass |
| this PR | **58%** (15/26) | ❌ `AssertionError: False != True` |

### Root cause

The #822 change is spec-correct:

```js
// lib/handlers/subscribe.js
this.granted = this.client.version === 5 ? ReasonCodes.NOT_AUTHORIZED : 128 // 0x87 for v5
```

`0x87` (Not authorized) is a valid, more-specific SUBACK reason code per **MQTT-5.0 §3.9.3** — strictly better than the coarse `0x80`.

But the compat harness (`tools/mqtt-compat/broker.js`) deliberately denies `test/nosubscribe` via `authorizeSubscribe` so Paho's `test_subscribe_failure` can run, and that Paho test hardcodes `assert value == 0x80` (written for 3.1.1, where `0x80` was the only failure code). So a spec **improvement** flips a previously-passing test to failing and drops MQTT 5.0 62% → 58%.

The report shows this as a plain ❌ failure, not an `⚠️ expected` gap, so it reads as a compatibility regression when it's actually a conformance improvement. The workflow only warns on *unexpected passes*, so nothing flagged the drop.

### Requested changes (same PR)

1. Add `test_subscribe_failure` (v5) to `EXPECTED_GAPS` in `tools/mqtt-compat/run_compat.py`, with a note along the lines of: *Paho asserts SUBACK `0x80`, but aedes correctly returns `0x87` (Not authorized) per MQTT-5.0 §3.9.3.* This restores the percentage and classifies it correctly — not a gap in aedes, but a Paho assertion that predates v5 granular reason codes.
2. Update the now-stale comment in `tools/mqtt-compat/broker.js` — it still says `deny -> SUBACK 0x80`, which is no longer true for v5 clients after this change.

Everything else (v3/v4 still `0x80`, MQTT 3.1.1 stays 100%, all functional CI green) looks good. `codecov/*` failures are coverage-threshold only.

BenjaminDobler:
Good catch on the compat-report regression — addressed in `afd1768` (also rebased onto the latest `mqttv5`).

- **`run_compat.py`:** added `test_subscribe_failure` to the v5 `EXPECTED_GAPS` with a note that Paho hardcodes `assert == 0x80` while aedes correctly returns `0x87` (Not authorized) per MQTT-5.0 §3.9.3 — classified as a Paho-side staleness, not an aedes gap, which restores the MQTT 5.0 %.
- **`broker.js`:** updated the stale `deny -> SUBACK 0x80` comment (v5 is now `0x87`; v3.1.1 stays `0x80`), with a pointer to the EXPECTED_GAPS entry.

The four earlier inline threads (no-PUBREL-on-rejecting-PUBREC, ackProperties comment, the two FYIs) were already fixed in `2f6f5bb` — feel free to resolve them.

BenjaminDobler:
Thanks for the Paho codec+test patch (`c3acc29`) — that's a cleaner fix than my `EXPECTED_GAPS` entry: it makes `test_subscribe_failure` actually pass and restores 62%, rather than masking it as a gap aedes doesn't have. 👍

The remaining red check was **codecov/patch** — the non-clean branch of the new `handlePubrec` (releasing the stored outgoing packet id via `outgoingClearMessageId` when a v5 PUBREC arrives with reason ≥ 0x80) was uncovered: my rejecting-PUBREC test only exercised a *clean* subscriber. Fixed in `5a7bd21` — parametrized the test to run both a clean and a non-clean (`sessionExpiryInterval`) subscriber. `pubrec.js` is now 100% and all other added lines are covered.

Rebased the stack on top: #1102 (`f643b8f`) and #1115 (`89ae10e`).

Issue #1091, opened 2026-06-29
Request / Response Information (CONNECT 0x19 → CONNACK 0x1A)


robertsLando:
Part of the MQTT 5.0 tracking issue #821.

## Feature

MQTT 5.0 **Request / Response Information**:

- **CONNECT** property `Request Response Information` (`0x19`, §3.1.2.11.6) — a one-byte flag the client sets to ask the server to return Response Information in the CONNACK.
- **CONNACK** property `Response Information` (`0x1A`, §3.2.2.3.15) — a UTF-8 string the server returns (typically a topic prefix / namespace) that the client uses as the basis for the Response Topic in the request/response pattern.

Normative rules: the server **MAY** return Response Information; if the client did **not** request it (flag `0` / absent), the server **MUST NOT** send Response Information ([MQTT-3.1.2-28]).

## Why it's separate from #827

This is **not** the same as #827 (Request/Response). #827 covers the per-PUBLISH application-level `Response Topic` + `Correlation Data`, which the broker relays verbatim. This issue is about the broker **generating** a response-topic namespace at connect time — a distinct property pair with its own MUST-NOT, and not covered by #832 (which is the boolean capability flags only).

## Current status

Not implemented (verified against the `mqttv5` branch / PR #1088 — no handling of `requestResponseInformation` / `responseInformation`). This is an **optional** server feature (server MAY), so omitting it is spec-conformant; tracked here for completeness toward full v5 support.

PR #1102, opened 2026-06-30
feat: MQTT 5.0 Request / Response Information (#1091) [stacked on #1100]

BenjaminDobler:
> 🔗 **Stacked on #1100 — please merge that first.**
> This branch is based on `feat/v5-ack-reason-codes` (PR #1100), so until #1100 merges the diff below also shows its commits. GitHub will narrow it to just this PR's changes automatically once #1100 lands. **Review only this PR's net changes here:** https://github.com/BenjaminDobler/aedes/compare/feat/v5-ack-reason-codes...feat/v5-request-response-info
>
> Kept as a **draft** until #1100 is merged; I'll rebase onto the updated `mqttv5`, force-push, and mark it ready.

PR 2 of the #821 follow-ups, per the [agreed sequence](https://github.com/moscajs/aedes/issues/821#issuecomment-4842940944). Broker-core only — no persistence changes, cluster-safe.

## #1091 — Request / Response Information
When a client sets **Request Response Information** on CONNECT (§3.1.2.11.6), the broker returns **Response Information** in the CONNACK (§3.2.2.3.15, server MAY) — the base string a client uses to build request/response topics.

- New broker option **`responseInformation`**: a static string **or** a `(client) => string | undefined` function (per-client / per-tenant base). Echoed only when the client requested it **and** the option is set; omitted otherwise (return `undefined` to opt out). Default `null`.
- `connect.js` parses `requestResponseInformation` (boolean; default `false`) and resolves the value in `doConnack` alongside the other CONNACK overrides (assigned client id / server keep alive / clamped session expiry).

## Tests
Returned when requested + configured; omitted when not requested; per-client function form.

## Docs / types
`responseInformation` option documented in `docs/Aedes.md` and typed on `AedesOptions`.

371 unit tests + lint + tsd green.


lib/handlers/connect.js
```diff
@@ -77,6 +77,17 @@ function connackProperties (broker) {
...
 }
 
+// [#1091] Resolve the broker's Response Information for this client. The
+// `responseInformation` broker option is either a static string or a
+// `(client) => string | undefined` function. Returns undefined (⇒ the property
+// is omitted) when not configured or the function opts out.
+function resolveResponseInformation (client) {
+  const ri = client.broker.responseInformation
+  if (ri == null) return undefined
+  const value = typeof ri === 'function' ? ri(client) : ri
```
robertsLando: **[Minor] · Correctness / robustness**

A user-supplied `responseInformation(client)` that **throws** propagates an uncaught synchronous exception: `resolveResponseInformation` → `doConnack` → `next` → `runSeries` (`utils.js` calls each action with no per-action try/catch), disrupting that client's CONNACK.

**Why:** the option's whole posture is tolerant — "return `undefined` to opt out" — yet a throw isn't contained. The static form can't throw, so this asymmetry is easy to hit (e.g. a per-tenant lookup that dereferences a field undefined for some clients). EMQX's equivalent is a static config string and cannot fail here.

**Fix:** wrap the resolve in try/catch and treat a throw as "omit" (same as `undefined`), so a per-tenant lookup failure degrades to no-property rather than a broken handshake.

test/mqtt5.js
```diff
@@ -611,6 +699,32 @@ test('MQTT 5.0 returns an Assigned Client Identifier for an empty clientId', asy
   t.assert.ok(broker.clients[assigned], 'client registered under the assigned id')
 })
 
+test('MQTT 5.0 returns Response Information when requested and configured', async (t) => {
```
robertsLando, line 722: **[Minor] · Tests**

Coverage is good but two spec-relevant branches are untested:

1. **Requested but broker not configured** (`responseInformation: null`) ⇒ omitted — the "configured gate" half of the only-when-requested-AND-configured rule.
2. **Function returns `undefined`** ⇒ omitted — the documented opt-out path.

The "not requested ⇒ omitted" case is covered; these two aren't. Both are one-liners mirroring the existing tests.

robertsLando (commented):
## Deep review — verdict: Ship ✅

Reviewed the net changes (vs `feat/v5-ack-reason-codes`). Focused verification of the three things called out: the mqtt-packet claims, MQTT 5.0 spec compliance, and how other brokers handle Response Information.

### mqtt-packet claims — legit (verified against mqtt-packet 9.0.2 source)
- **Decode:** `parser.js` `case 'byte': return this._parseByte() !== 0` — byte properties decode to a **boolean**, so `packet.properties?.requestResponseInformation === true` is exactly right (wire `1`→`true`, `0`→`false`, absent→`undefined`→`false`). Using `=== true` rather than the sibling's `!== false` is the correct call because the spec default here is **false**.
- **Encode:** `responseInformation` is a known property (id **26**, type **`string`** in `constants.js`); `writeToStream.js` `getProperties()` serializes any present property generically — there is **no per-packet-type whitelist** — so it is emitted on CONNACK as a UTF-8 string, and a non-string value is caught (`stream.destroy('Invalid responseInformation…')`) rather than producing a corrupt frame.

### MQTT 5.0 spec — compliant (verified against OASIS mqtt-v5.0)
- Section refs exact: **§3.1.2.11.6** Request Response Information (id 25/0x19, Byte) and **§3.2.2.3.15** Response Information (id 26/0x1A, UTF-8 String) — all IDs/types/titles match Table 2-4.
- **[MQTT-3.1.2-28]** (value 0 ⇒ server MUST NOT return) is honored: the `=== true` gate means absent/0 is never returned. Server-MAY is honored (returned only when also configured; `null` default ⇒ never). Only-on-request is spec-mandated and correctly implemented.
- Content/format is application-defined per spec, so treating the value as opaque (no slash/prefix manipulation) is correct.

### Other brokers — design aligns / is a superset
- **Mosquitto:** the broker does not implement this at all — `RESPONSE_INFORMATION` exists only in the client lib + tests; `handle_connect.c` never adds it to CONNACK, and there is no config option.
- **EMQX:** built-in `mqtt.response_information`, **static string only**, only-on-request, empty ⇒ omit. Aedes's static-string branch matches this exactly.
- **HiveMQ / VerneMQ:** per-client values only via extension/hook code. Aedes folds that per-client pattern into a first-class `(client) => string | undefined` option — the primary use case the spec/EMQX cite (client-scoped prefix for ACL alignment).

### Strengths
Accurate, well-cited comments; correct default handling; v5-only + only-on-request gating; string|function option is a genuine superset of other brokers; clean integration into the existing `doConnack` override block.

### FYI (non-blocking)
- **Empty string is sent, not omitted:** `resolveResponseInformation` only treats `== null` as omit, so `responseInformation: ''` emits a zero-length Response Information (spec-valid, but the docs frame omission as `undefined`). Confirm it's deliberate.
- **Per-client timing is correct:** `client.id` is assigned in `init()` before `runSeries`, so `resp/${client.id}` resolves the final id (including the auto-assigned UUID for empty-clientId clients).
- **No trailing-slash manipulation** — matches every broker surveyed; keep the value opaque as you do.

Two Minor items threaded inline. Nothing blocking.

BenjaminDobler:
Thanks — ship-blocking nothing, but addressed both Minors in `dcfc175`:

- **Throwing resolver:** `resolveResponseInformation` now wraps the option in try/catch and treats a throw as "omit" (same as `undefined`), so a per-tenant lookup that throws degrades to no-property instead of breaking the CONNACK.
- **Tests:** added the two missing branches — requested-but-not-configured (`responseInformation: null`) ⇒ omitted, and a function returning `undefined` ⇒ omitted — plus a throwing resolver ⇒ omitted with the handshake intact.
- **FYI (empty string):** deliberate — `''` is a spec-valid Response Information, so only `undefined`/`null` omit; the docs say "return `undefined` … to omit" for exactly that reason.

(Also rebased onto the latest `mqttv5` via the #1100 base.)

robertsLando (commented):
## Deep review — Request / Response Information

**Verdict: Ship with minor changes.** No blockers, no confirmed majors on the core feature. Core gates verified spec-conformant (§3.1.2.11.6 / §3.2.2.3.15): packet-id lifecycle, SUBACK `0x87` exclusion, and `_requestProblemInformation` gating all sound. The 2 prior review threads are genuinely addressed (throwing `responseInformation` resolver now `try/catch` + `== null` opt-out; both previously-untested branches now covered in `test/mqtt5.js`) — resolved.

### Worth addressing
- **Observability (`lib/handlers/connect.js:92`)** — the empty `catch {}` in `resolveResponseInformation` silently swallows a throwing per-client resolver with zero telemetry. A resolver that throws for every client never emits the property, forever, undiagnosable. The same file already surfaces swallowed failures via `client.emit('error', err)`; suggest emitting there before returning `undefined`.
- **CI gating (`.github/workflows/mqtt-compat.yml`)** — `run_compat.py` never exits non-zero on a compat regression, so the job is advisory-only by construction: a change cratering v5 compat still shows green. Either document it as informational, or add a no-regression / min-percentage assertion.

### Minor
- `test/types/aedes.test-d.ts:28` — the `responseInformation` union only exercises the function arm; the static-string arm (the common config) is never type-checked. One-line fix.
- `types/instance.d.ts` — the union omits `null`, though `null` is the documented disable sentinel, so `responseInformation: null` is currently a type error.

### Nits
- `docs/Aedes.md` — the throw-degrades-to-omit contract isn't documented (only undefined/unset is).
- `0x80` reason-code threshold appears as a bare literal in both `pubrec.js` and `subscribe.js` — consider a named constant.
- `.gitignore` — ignores `__pycache__/`/`*.pyc` but not the harness-generated `paho/`, `report.md`, `result.json` at repo root (a `git add -A` footgun for local reproducers).
- `lib/handlers/pubrec.js` (~17–24) — the rejecting-PUBREC path frees the packet id but emits no failure/`'ack'` event, so an app tracking QoS 2 completion sees the message silently vanish. Packet-id invariant holds; consider a terminal event or doc note.
- Security — `ackProperties(client, err.message)` reflects the app's `authorizePublish` error message verbatim to an untrusted publisher (an ACL-probing oracle). Framework handling is correct (v5-only, RPI-gated) and `docs/Aedes.md:78` already warns; consider making the warning more prominent. No code change.

_Note: this branch is CONFLICTING against `mqttv5` (expected for the draft stack) and will need a rebase once #1100 lands._

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

BenjaminDobler:
Thanks — rebased onto `mqttv5` (post-#1100) and addressed the review in `3bce1b4`:

**Minor**
- `types/instance.d.ts`: `responseInformation` union now includes `null` (the documented disable default was a type error). tsd now exercises the static-string and `null` arms too.

**Worth addressing**
- **Observability:** a throwing resolver is no longer swallowed by an empty `catch{}` — it's surfaced on the broker `clientError` event. Note: I used `client.broker.emit('clientError', client, err)` rather than the suggested `client.emit('error', err)` — the latter routes through `_onError`, which destroys the connection, and this runs mid-`doConnack`, so it would break the very CONNACK we're building. Property is still omitted; test asserts the event + intact handshake.

**Nits**
- Docs: documented the throw-degrades-to-omit + `clientError` contract.
- Extracted `REASON_CODE_ERROR_THRESHOLD = 0x80` (constants.js) and used it in `pubrec.js` / `subscribe.js`.
- `.gitignore`: added the harness's local artifacts (`paho/`, `report.md`, `result.json`).

**Deferred (with rationale)**
- **CI gating** (`run_compat.py` advisory-only): that's your compat tooling — happy to add a min-% / no-regression gate if you want, but changing the job's gating semantics felt like your call rather than something to slip into this feature PR.
- **pubrec rejecting-PUBREC emits no terminal `ack`** and **making the `authorizePublish` error-message-visible warning more prominent**: both touch now-merged #1100 code and are 'consider' nits; happy to do a small follow-up PR against `mqttv5` for them if you'd prefer.

robertsLando (approved):

PR #1115, opened 2026-07-02
feat: MQTT 5.0 Server Reference / connect-time redirect (#838) [stacked on #1102]

BenjaminDobler:
> 🔗 **Stacked on #1102 → #1100.** Merge order: **#1100 → #1102 → this**.
> This branch is based on `feat/v5-request-response-info` (PR #1102, itself on #1100), so until those merge the diff below also shows their commits. GitHub narrows it automatically as each parent lands. **Review only this PR's net changes here:** https://github.com/BenjaminDobler/aedes/compare/feat/v5-request-response-info...feat/v5-server-reference
>
> Kept as a **draft** until #1100 and #1102 are merged; I'll then rebase onto the updated `mqttv5`, force-push, and mark it ready.

PR 3 of the #821 follow-ups, per the [agreed sequence](https://github.com/moscajs/aedes/issues/821#issuecomment-4842940944). Broker-core only — no persistence changes, cluster-safe.

## #838 — Server Reference / connect-time redirect (§4.11)
Lets the broker redirect a v5 client to another server.

- **Connect-time:** `preConnect` or `authenticate` redirect by rejecting with an error carrying a `serverReference` (and optional `reasonCode`). The rejection CONNACK then carries reason code **`0x9C`** (Use another server) — or **`0x9D`** (Server moved) when the hook sets `err.reasonCode` — plus the `serverReference` property. Per the [chosen design](https://github.com/moscajs/aedes/issues/821): reuse the existing hooks' error, no new API surface.
- **Mid-session:** already works via `client.disconnect({ reasonCode, properties: { serverReference } })` — covered by a test to lock it in.

- `lib/constants.js`: `ReasonCodes.USE_ANOTHER_SERVER` (`0x9C`) / `SERVER_MOVED` (`0x9D`).
- `connect.js`: `rejectConnect` now forwards CONNACK `properties`; the `preConnect` negate path keeps `connecting` true so `write()` still emits the redirect CONNACK before closing.

## Tests
`preConnect` redirect → `0x9C` + serverReference; `authenticate` redirect with a custom `0x9D`; server-initiated DISCONNECT carrying a serverReference.

## Docs
`preConnect` handler redirect note (docs/Aedes.md) + a new `client.disconnect()` section (docs/Client.md).

374 unit tests + lint + tsd green.


docs/Aedes.md
```diff
@@ -332,6 +333,16 @@ Invoked when server receives a valid [`CONNECT`][CONNECT] packet. The packet can
 
 Any `error` will be raised in `connectionError` event.
 
+> __MQTT 5.0 server redirect:__ to redirect a v5 client to another server, reject the connection with an `error` carrying a `serverReference` (and optionally a `reasonCode`). The rejection CONNACK then carries reason code `0x9C` (Use another server; or `0x9D` Server moved when `error.reasonCode` is set to it) and the `serverReference` property. The same works from [`authenticate`](#handler-authenticate-client-username-password-callback), and mid-session via [`client.disconnect`](./Client.md#clientdisconnect-opts-callback). §4.11
```
robertsLando: **[Minor] · Design/Docs**

The examples show a single reference (`other-host.example:1883`), which reads as "one host only." Per MQTT 5.0 §4.11 the Server Reference is normatively **a space-separated list of references**; `host:port` per element is only non-normative guidance (IPv6 literals bracketed, e.g. `[fe80::1]:1883`). Since aedes passes the hook's string through verbatim, multi-server already works.

**Why:** a spec-literate reader would assume single-host-only and miss that load-balancing across several targets is already supported.

**Fix:** one line noting that `serverReference` may be a space-separated list of `host[:port]` references. (Also worth a docstring pointer that the DISCONNECT Server Reference property lives in spec §3.14.2.2.5.)

lib/handlers/connect.js
```diff
@@ -114,10 +125,28 @@ function handleConnect (client, packet, done) {
...
+    // [#838] Connect-time server redirect: a preConnect error carrying a
+    // serverReference redirects a v5 client to another server via a CONNACK
+    // (default 0x9C Use another server; 0x9D Server moved via err.reasonCode).
+    // client.version isn't assigned before init(), so key off the CONNECT.
+    // Keep `connecting` true here so write() still emits the CONNACK; clear it
+    // once the packet is on the wire.
+    if (err && err.serverReference && packet.protocolVersion === 5) {
+      rejectConnect(client, {
+        error: err,
+        reasonCode: err.reasonCode ?? ReasonCodes.USE_ANOTHER_SERVER,
```
robertsLando: **[FYI] · Correctness/Spec**

`reasonCode: err.reasonCode ?? ReasonCodes.USE_ANOTHER_SERVER` forwards *any* `err.reasonCode` the hook sets (same in the `authenticate` path, ~line 349). A hook pairing `serverReference` with a non-redirect code (e.g. `0x97` Quota exceeded) would emit a reference that the spec binds only to `0x9C`/`0x9D`.

**Why:** not a bug — it's opt-in and the hook author's responsibility — but the pairing is easy to get subtly wrong.

**Fix:** optional. Either a doc line ("`serverReference` is only meaningful with `0x9C`/`0x9D`") or clamp non-redirect codes when a `serverReference` is present.

test/mqtt5.js
```diff
@@ -611,6 +699,83 @@ test('MQTT 5.0 returns an Assigned Client Identifier for an empty clientId', asy
...
+})
+
+test('MQTT 5.0 server-initiated DISCONNECT can carry a Server Reference', async (t) => {
+  t.plan(2)
+  const { broker, connect } = await createServerAndConnect(t)
+  const client = connect({ clientId: 'redir-disc', reconnectPeriod: 0 })
+  await once(client, 'connect')
+  const disc = once(client, 'disconnect')
+  // Poll for broker-side connected (set after the CONNACK is written).
+  while (!broker.clients['redir-disc']?.connected) await delay(5)
```
robertsLando: **[Nit] · Tests**

Busy-poll on broker-side readiness. `await once(broker, 'clientReady')` (or awaiting the `connackSent`/`clientReady` signal) would be less timing-fragile than a 5 ms spin loop.

**Why:** minor test robustness; the inline comment already acknowledges the poll.

**Fix:** optional — swap the loop for the event.

robertsLando (commented):
## Deep review — MQTT 5.0 Server Reference / connect-time redirect

**Verdict: Ship.** The approach is sound, the `mqtt-packet` claims check out, and the wire behavior matches the MQTT 5.0 spec. Findings below are docs/nits only — no blockers.

### Verification (the two things worth being sure about)

**`mqtt-packet` support is real.** `mqtt-packet@9.0.2` supports `serverReference` as property id **28 (0x1C)**, type `string`, and its own test vectors exercise it on **both** CONNACK and DISCONNECT. The full serialization path holds: `write.js` passes `protocolVersion: 5` into `writeToStream`, and the v5 `Connack` carries `reasonCode` + `properties`, so `properties.serverReference` genuinely reaches the wire.

**Spec compliance confirmed** against the OASIS MQTT 5.0 standard:
- `0x9C` "Use another server" (temporary) / `0x9D` "Server moved" (permanent) — names/values correct, and **both are valid in CONNACK *and* DISCONNECT** (Table 2-6, §3.2.2.2, §3.14.2.1).
- Server Reference property `0x1C`, UTF-8 string, allowed in CONNACK (§3.2.2.3.16) and DISCONNECT (§3.14.2.2.5).
- §4.11 semantics honored: Server **MAY** include a reference, Client **SHOULD** switch. Reusing the hook `error` to carry it fits the §821 design with zero new API surface.

### How other brokers handle this (for context)

| Broker | Connect-time (CONNACK) redirect | Mid-session (DISCONNECT) redirect |
|---|---|---|
| **Mosquitto** | none | none — man page: *"Mosquitto does not currently make use of this feature."* |
| **EMQX** | none | node-evacuation `rebalance … --redirect-to "H1:P1 H2:P2"` (space-separated) |
| **HiveMQ** | none | reason-code enums exist, but the documented disconnect API exposes no Server Reference |
| **aedes (this PR)** | ✅ `preConnect`/`authenticate` | ✅ `client.disconnect()` |

aedes is actually ahead here: none of the three major brokers offer a connect-time CONNACK redirect hook.

### Strengths
- The `connecting`-stays-true handling is correct: `write()` gates on `waitForDrain`, so the CONNACK is flushed before `connecting=false` → close — no truncation even under backpressure.
- Good comments on the non-obvious `packet.protocolVersion` (preConnect) vs `client.version` (authenticate) distinction.

### One doc-precision note
The DISCONNECT Server Reference section is **§3.14.2.2.5** (not .15), and per §4.11 the Server Reference is normatively a **space-separated list of references** — `host:port` per element is only non-normative guidance. See the inline comment on `docs/Aedes.md`.

Inline findings follow. Nothing blocking; all optional.

BenjaminDobler:
Thanks — docs/nits addressed in `9b27a8c`:

- **§4.11 space-separated list:** the docs now note `serverReference` may be a space-separated list of `host[:port]` references (IPv6 bracketed), passed through verbatim, so multi-target load-balancing is discoverable; the example shows two.
- **Spec ref:** corrected the DISCONNECT Server Reference reference to §3.14.2.2.5.
- **FYI (reason-code pairing):** added a doc line that `serverReference` is only meaningful paired with `0x9C`/`0x9D` (kept it a doc note rather than clamping, since it's the hook author's call).
- **Test nit:** the DISCONNECT-redirect test now awaits the broker `clientReady` event instead of busy-polling.

(Also rebased onto the latest `mqttv5` through the stack.)

robertsLando (commented):
## Deep review — Server Reference / connect-time redirect

**Verdict: Ship.** No blockers, no majors. The approach is sound and reuses the existing `rejectConnect` / `doConnack` path cleanly.

Verified at head:
- v5 gating is correct — `negate` (preConnect) keys off `packet.protocolVersion === 5`, `authenticate` keys off `client.version === 5`; no NPE on the redirect branch (`err` is always constructed by that point).
- `rejectConnect` now always passes `properties` (possibly `undefined`); `doConnack` guards it, so existing reject callers don't regress and capability advertising is correctly skipped on redirects.
- The 3 prior review threads are all genuinely addressed (docs: space-separated `host[:port]` list + IPv6 bracketing + spec cite; reasonCode doc note; test busy-poll → `once(broker, 'clientReady')`) — resolved.

### Nits (non-blocking)
- `lib/handlers/connect.js` (~145, ~355): `err.reasonCode ?? USE_ANOTHER_SERVER` lets a hook pair `reasonCode: 0x00` **with** a `serverReference`, emitting a success-coded CONNACK on a rejection path. Opt-in footgun, covered by the docs; clamping to `0x9C`/`0x9D`-only would harden it.
- Test coverage: no negative test that a v3/v4 client whose reject carries `serverReference` is **not** redirected (the property must be silently dropped), and no assertion the socket closes post-redirect. Cheap to lock down.

_Note: this branch is CONFLICTING against `mqttv5` and will need a rebase once the stack (#1100 → #1102) lands._

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

BenjaminDobler:
Thanks — rebased onto the updated stack and addressed both nits in `9e6835f`:

- **Reason-code clamp:** extracted `redirectReasonCode()` and applied it on both the `preConnect` and `authenticate` redirect paths. A hook pairing `serverReference` with a non-redirect code (e.g. `0x00`/`0x97`) is now clamped to `0x9C` (`0x9D` still honored), so a rejection can't emit a success-coded CONNACK. Docs updated to describe the clamp.
- **Tests:** a non-redirect `reasonCode` is clamped to `0x9C`; a non-v5 (v4) client whose reject carries `serverReference` is **not** redirected (the property is dropped); and the socket **closes** after the redirect CONNACK.

Still stacked on #1102 (draft until that lands).

robertsLando:
@BenjaminDobler please rebase this one then I will do a final review

BenjaminDobler:
Rebased onto the latest `mqttv5` (`18dd139`, now includes the merged #1100 + #1102) — conflict-free, no longer showing as behind. lint + tsd + full unit (367) green. Ready for your final review 🙏

lib/handlers/connect.js
```diff
@@ -336,6 +364,14 @@ function authenticate (arg, done) {
       arg.returnCode = 5
       err = new Error(errorMessages[arg.returnCode])
     }
+    // [#838] Server redirect: an authenticate error carrying a serverReference
+    // redirects a v5 client to another server via the CONNACK — default 0x9C
+    // (Use another server); the hook may set 0x9D (Server moved) via
+    // err.reasonCode. §4.11
+    if (client.version === 5 && err.serverReference) {
```
robertsLando, line 371: **[Minor] · Observability**

The `authenticate` redirect path sets `arg.reasonCode`/`arg.properties` but never attaches the redirect code to the emitted `err` (it keeps `err.errorCode = arg.returnCode`, i.e. 4/5). The `preConnect` path *does* get `err.reasonCode = 0x9C/0x9D` via `rejectConnect`.

**Why:** an operator watching `connectionError` sees the redirect reason on preConnect redirects but not authenticate ones — and `rejectConnect`'s own comment advertises that `reasonCode` is attached to the error "so a reject storm is diagnosable from the error alone." This path diverges from that contract.

**Fix:** consider `err.reasonCode = arg.reasonCode` inside this block to align the two hooks. (Author's call — it also follows the existing `authenticate`-uses-`errorCode` convention, so leaving it is defensible.)

robertsLando (commented):
Deep review — MQTT 5.0 §4.11 spec-compliance pass. Approach is sound and spec-accurate: 0x9C/0x9D reason codes, `sessionPresent=0` on non-zero CONNACK code [MQTT-3.2.2-4], ≥0x80 closes the connection [MQTT-3.2.2-7], v5-gated so v3/v4 never leak a `serverReference`, and the docs' "space-separated list of `host[:port]`" claim matches §4.11 (verified against the OASIS spec). No blockers. A few optional nits + test-coverage notes inline.

lib/constants.js
```diff
@@ -36,6 +36,10 @@ export const ReasonCodes = {
   RECEIVE_MAXIMUM_EXCEEDED: 0x93,
   TOPIC_ALIAS_INVALID: 0x94,
   PACKET_TOO_LARGE: 0x95,
+  // Server redirect (CONNACK / DISCONNECT): try a different server now, or the
+  // client's session has been moved permanently. [MQTT-3.14.2.1] §4.11
```
robertsLando: **[Nit] · Spec reference**

`[MQTT-3.14.2.1] §4.11` reads like a normative-statement ID but is a section pointer, and it points only at DISCONNECT. These codes are equally used in **CONNACK** (§3.2.2.2 reason codes / §3.2.2.3 Server Reference) — which is in fact this PR's primary path.

**Fix:** e.g. `§4.11 (CONNACK §3.2.2, DISCONNECT §3.14.2)`.

test/mqtt5.js
```diff
@@ -781,6 +781,98 @@ test('MQTT 5.0 a Response Information function returning undefined (or throwing)
...
+      authenticate: (client, username, password, cb) => {
+        cb(Object.assign(new Error('moved'), { serverReference: 'new-host:1883', reasonCode: 0x9D }), false)
+      }
+    }
+  })
+  const client = connect({ clientId: 'redir-auth', reconnectPeriod: 0 })
+  client.on('error', () => {})
+  const [connack] = await once(client, 'packetreceive')
+  t.assert.equal(connack.reasonCode, 0x9D, '0x9D Server moved (custom reason code)')
+  t.assert.equal(connack.properties?.serverReference, 'new-host:1883', 'server reference on the wire')
```
robertsLando, line 862: **[Minor] · Test coverage**

This authenticate-redirect test asserts the CONNACK reason code + `serverReference` but not that the connection then **closes** — whereas the preConnect test does assert `client.connected === false`. Since [MQTT-3.2.2-7] requires a ≥0x80 CONNACK to close, and the two paths use different closure mechanisms (`client.close.bind` here vs `done(err)` there), an explicit close assertion here would lock the authenticate path's closure independently.

test/mqtt5.js
```diff
@@ -781,6 +781,98 @@ test('MQTT 5.0 a Response Information function returning undefined (or throwing)
...
+        cb(Object.assign(new Error('go elsewhere'), { serverReference: 'other-host:1883' }), false)
+      }
+    }
+  })
+  const client = connect({ clientId: 'redir-pc', reconnectPeriod: 0 })
+  client.on('error', () => {}) // a >= 0x80 CONNACK surfaces as a client error
+  // Plain close promise: events.once(…, 'close') would reject on the 0x9C 'error'.
+  const closed = new Promise(resolve => client.once('close', resolve))
+  // The CONNACK is the first packet the broker sends.
+  const [connack] = await once(client, 'packetreceive')
```
robertsLando, line 802: **[Minor] · Test coverage**

No test asserts the redirect CONNACK carries **only** `serverReference` and not the broker-capability properties (`receiveMaximum`, `topicAliasMaximum`, …). It's correct today because `doConnack`'s success-guard is skipped once `reasonCode` is set — but that's exactly the kind of invariant worth pinning against regression. Consider asserting the absence of a capability property here (e.g. `connack.properties?.receiveMaximum === undefined`).

BenjaminDobler:
Thanks — addressed all four nits in `a4f5608` (all optional, but cheap and worth it):

- **Observability:** the `authenticate` redirect path now also sets `err.reasonCode` (0x9C/0x9D), matching the `preConnect`/`rejectConnect` contract so the redirect reason shows up on `connectionError`, not just the CONNACK.
- **Spec ref:** the constant's comment now cites CONNACK §3.2.2 and DISCONNECT §3.14.2 (both paths) instead of the DISCONNECT-only `[MQTT-3.14.2.1]` pointer.
- **Test — capability omission:** the preConnect-redirect test now advertises `receiveMaximum`/`topicAliasMaximum` and asserts the redirect CONNACK carries **neither** (pins the `doConnack` success-guard invariant).
- **Test — closure:** the authenticate-redirect test now asserts the connection closes too [MQTT-3.2.2-7], locking that path's closure independently of preConnect.

lint + tsd + full unit green.

robertsLando (approved):

PR #1116, opened 2026-07-06
feat: MQTT 5.0 Reason String on the remaining ACKs (#823) + PUBCOMP 0x92 (#822)

BenjaminDobler:
First of the next #821 batch, per the [agreed sequence](https://github.com/moscajs/aedes/issues/821#issuecomment-4893127917) — the low-blast-radius conformance win you flagged (#823 is only partial after #1100). Broker-core only, cluster-safe, no persistence.

## What

#1100 landed the Reason String on PUBACK/PUBREC; this completes the remaining ACK types, all routed through the same `ackProperties` gate (v5 + Request Problem Information):

- **PUBCOMP (#822 + #823):** when a PUBREL references an unknown packet id, answer with reason code **`0x92`** (Packet Identifier not found) instead of `0x00`, plus a Reason String [MQTT-3.7.2.1]. Normal QoS 2 flows (packet found) stay `0x00`; v3/v4 unaffected.
- **SUBACK (#823):** a failed subscription now carries a broker-generated Reason String — `0x9E` → "shared subscriptions are not supported", `0x87` → "not authorized to subscribe", else "subscription refused".
- **UNSUBACK (#823):** "no subscription existed for one or more topic filters" on `0x11`.

## Design note

Reason strings are **broker-generated** (no new hook API), matching #1100's "surface what the broker knows" posture. The `authorizeSubscribe` deny path (`callback(null, null)`) has no message channel, so the SUBACK string is derived from the reason code. If you'd prefer letting `authorizeSubscribe` supply a custom string, that's a small follow-up — happy to add it.

**PUBREL** stays success-only: aedes doesn't detect a not-found on the outbound QoS 2 ack path (would need an extra outgoing-store lookup on the hot path), so there's no non-`0x00` case to annotate there. Flagged for a possible follow-up.

## Tests
SUBACK reason string (authz-deny + `$share`), UNSUBACK reason string, PUBCOMP `0x92` + reason string (raw v5 PUBREL for an unknown packet id).

369 unit tests + lint + tsd green.


robertsLando (commented):
Deep review — approach is sound and MQTT 5-conformant. All three ACKs route through the existing `ackProperties` gate (v5 + non-empty string + Request Problem Information), reason codes are valid for their packet types (PUBCOMP `0x92` [3.7.2.1], SUBACK `0x87`/`0x9E` [3.9.3], UNSUBACK `0x11` [3.11.3]), the RPI default of 1 is honored, and no v5 properties are clobbered. Two non-blocking notes inline. **Verdict: ship with minor changes.**

lib/handlers/pubrel.js
```diff
@@ -10,11 +10,20 @@ class ClientPacketStatus {
...
+  constructor (packet, client, foundInStore) {
     this.cmd = 'pubcomp'
     this.messageId = packet.messageId
-    // [#822] MQTT 5.0 success reason code (0x00); ignored when serializing v3/v4.
-    this.reasonCode = ReasonCodes.SUCCESS
+    // MQTT 5.0 reason code: 0x00 on success, or 0x92 (Packet Identifier not
+    // found) when the PUBREL referenced an unknown packet id — with a Reason
+    // String [#822/#823], gated by Request Problem Information. Ignored when
+    // serializing v3/v4. [MQTT-3.7.2.1]
+    if (client.version === 5 && !foundInStore) {
```
robertsLando, line 20: **[Minor] · Correctness** — a transient persistence read failure is conflated with "packet not found."

`pubrelGet`'s reject handler sets `arg.foundInStore = false` (unchanged, just below the diff), and this line now feeds that flag into the client-visible reason code. So a DB hiccup on `incomingGetPacket` makes the broker answer `0x92 Packet Identifier not found` for an id that may well exist. That `foundInStore = false` on error was originally chosen for the *delete* decision (safe: don't delete what you couldn't confirm) — reusing it here gives it a new, outward-facing meaning.

**Why:** `0x92` can prompt a conformant client to log an error or treat its session state as mismatched, over a purely internal, recoverable fault; before this PR that case returned a benign `0x00`.

**Fix:** distinguish lookup-failure from a confirmed miss — only emit `0x92` when the lookup *succeeded and returned nothing*, and fall back to `0x00` on the persistence error path.

lib/handlers/subscribe.js
```diff
@@ -16,12 +16,32 @@ const restoreTopicActions = runFall([
...
-  constructor (packet, granted) {
+  constructor (packet, granted, client) {
     this.cmd = 'suback'
     this.messageId = packet.messageId
     // the qos granted
     this.granted = granted
+    // [#823] Reason String on a failed subscription, gated by Request Problem
+    // Information.
+    const properties = ackProperties(client, subackReasonString(granted))
+    if (properties) this.properties = properties
```
robertsLando, line 44: **[FYI] · Spec** — none of the three ACKs check the client's `Maximum Packet Size` before appending a Reason String [MQTT-3.1.2-24].

Not a defect here: the strings are short and fixed, and SUBACK/UNSUBACK/PUBCOMP are tiny packets — so the limit is realistically never hit. It's also the same pattern #1100 already shipped for PUBACK. Flagging only so it's a conscious, tracked assumption rather than an oversight, in case Reason Strings ever grow or User Properties get added to this path later.

lib/handlers/pubrel.js
BenjaminDobler: Good catch on the conflation — but digging in, the suggested fix doesn't hold with the persistence contract, and I'd like your call before changing it.

`incomingGetPacket` does **not** resolve-with-nothing on a miss: aedes-persistence **throws** `Error('no such packet')` (`asyncPersistence.js:291-293`). So "packet not found" and "transient read error" arrive on the *same* rejection path — and the `0x92` feature (#822) is built on exactly that reject path. "Fall back to `0x00` on the persistence error path" would therefore delete the whole feature (the unknown-id case *is* the error path), and it breaks the existing `0x92` test.

So the two cases aren't distinguishable at this layer without a discriminator. Options:

1. **Leave as-is.** The bundled aedes-persistence is in-memory and rejects `incomingGetPacket` *only* for not-found (no transient faults possible), so `0x92`-on-reject is correct for the default. An external persistence that rejects on a transient fault would emit `0x92` — an accepted edge, worth a doc note.
2. **Discriminate on a sentinel** — treat only `err.message === 'no such packet'` as a confirmed miss (`0x92`), else `0x00`. Fixes the transient case for the default persistence but couples to its error string and would regress `0x92` for a custom persistence that signals not-found differently.
3. **Persistence-contract change** — add a distinct not-found signal (resolve-`undefined` vs reject-on-error) in aedes-persistence and key off it. Cleanest, but spans repos.

My preference is **(1) + a doc note**, or defer **(3)** to a tracked issue. Happy to implement whichever you'd like — reverted my WIP for now.

lib/handlers/subscribe.js
BenjaminDobler: Agreed — conscious assumption, not an oversight (short fixed strings, tiny packets, same pattern #1100 shipped for PUBACK). Tracked it in #1125 so it's revisited if Reason Strings grow or User Properties get added to this path.

robertsLando (approved):
Sorry I'm super busy lately, let's merge this one, ignore the minor

Issue #1125, opened 2026-07-13
MQTT 5.0: honor Maximum Packet Size before appending Reason String / User Properties on ACKs


BenjaminDobler:
Tracked from the #1116 review (FYI thread on `lib/handlers/subscribe.js`).

None of the MQTT 5.0 ACKs — PUBACK (shipped in #1100), and PUBREC/PUBCOMP/SUBACK/UNSUBACK (#1116) — check the client's advertised `Maximum Packet Size` [MQTT-3.1.2-24] before appending a Reason String via `ackProperties`.

Today this is safe: the reason strings are short, fixed, and these are tiny control packets, so the limit is realistically never reached. Filing so the assumption is tracked rather than an oversight — it would matter if Reason Strings grow, or if User Properties are ever added to this path.

Fix (when needed): in `ackProperties` (lib/utils.js), drop the Reason String (and any future User Properties) when the resulting packet would exceed `client.maximumPacketSize`.

Issue #1124, opened 2026-07-13
MQTT 5.0: add a broker-side cap for outbound Topic Alias allocation


BenjaminDobler:
Follow-up from the review of #1117 (outbound broker-assigned Topic Alias).

Today the per-connection outbound alias table (`client._outboundTopicAliases`) is bounded only by the value the **client** advertises in its CONNECT `Topic Alias Maximum` (≤ 65535) × the number of distinct topics actually delivered. The inbound direction has a broker-side ceiling (the `topicAliasMaximum` option), but the outbound direction has no symmetric broker cap — an untrusted client can advertise `65535`, subscribe to `#`, and drive its per-connection `Map<topic, alias>` toward 65535 entries. It's self-inflicted and bounded per connection, but across many connections it's a memory-amplification vector, and it's asymmetric with the inbound path.

### Proposal
Add a broker option (e.g. `outboundTopicAliasMaximum`, default `0` = no broker cap, preserving current behaviour) and clamp the effective per-connection outbound maximum to `min(clientAdvertised, brokerCap)` in the connect handler — natural symmetry with the existing inbound `topicAliasMaximum` option.

Raised in the #1117 review thread on `lib/handlers/connect.js`; deferring to keep that PR focused (the feature is correct and safe as-is; this only adds an opt-in ceiling).

robertsLando:
closed in #1117

PR #1119, opened 2026-07-07
feat: MQTT 5.0 Enhanced Authentication (§4.12) — AUTH-packet exchange

BenjaminDobler:
Implements MQTT 5.0 Enhanced Authentication (§4.12) — the `AUTH`-packet challenge/response — via a new `authenticateEnhanced` broker hook. Closes the initial-auth part of #833; re-authentication is a fast-follow (see below).

Built on the prototype we validated in the [#833 discussion](https://github.com/moscajs/aedes/issues/833) (§1 pre-CONNACK dispatch, §2 pipeline suspension, §5 CONNACK-property merge), now fleshed out with the timer, error paths, types and docs.

### API

```js
new Aedes({
  authenticateEnhanced (client, method, data, done) {
    if (method !== 'SCRAM-SHA-256') {
      return done(Object.assign(new Error('bad method'), { reasonCode: 0x8C }))
    }
    if (data.toString() === 'client-final') done(null, { done: true, data: serverFinal })
    else done(null, { done: false, data: challenge }) // send AUTH 0x18, await client reply
  }
})
```

- Default `null` → a CONNECT asking for enhanced auth is rejected with **0x8C** (Bad authentication method). This is gated on the missing hook; auth-data-without-method stays an unconditional **0x82** regardless.
- `done(null, { done: false, data, properties })` → challenge (server `AUTH 0x18`).
- `done(null, { done: true, data })` → accept; `data` becomes the CONNACK Authentication Data.
- `done(error)` → reject; CONNACK carries `error.reasonCode` (default 0x87) + optional `error.reasonString`.

### How it works

- **Dispatch (§1)** — `client.js` `enqueue` now dispatches an `AUTH` immediately while `client.connecting`, so the continuation `AUTH` isn't queued behind a `'connected'` that can't fire before CONNACK.
- **Suspension (§2)** — the connect pipeline's `authenticate` step starts the exchange and does **not** resolve; `auth.js` `pause()`s the read loop across each hook call, writes `AUTH 0x18` + `resume()`s on continue, and invokes a connect-supplied `onSuccess`/`onFailure` closure on completion (no circular import into `doConnack`).
- **CONNACK merge (§5)** — final Authentication Data is **merged into** the generated CONNACK properties, so `sharedSubscriptionAvailable` etc. are still advertised.
- **Timer (§6)** — each round is bounded by `connectTimeout`; a stalled exchange (client never answers a challenge) is closed.
[...]

PR #1117, opened 2026-07-06
feat: MQTT 5.0 outbound (broker-assigned) Topic Alias (#840)

BenjaminDobler:
Second of the next #821 batch, per the [agreed sequence](https://github.com/moscajs/aedes/issues/821#issuecomment-4893127917) — the other low-blast-radius win. Broker-core only, cluster-safe, no persistence. Flips `test_server_topic_alias` from an `EXPECTED_GAP` toward a pass.

## What

Inbound topic aliases (client→broker) already work; this adds the reverse (broker→client). When a client advertises a **Topic Alias Maximum > 0** in its CONNECT (§3.1.2.11.5), the broker assigns aliases on outbound PUBLISH (§3.3.2.3.4):

- first PUBLISH on a topic → **full topic + Topic Alias** (registers the mapping)
- later PUBLISHes on that topic → **empty topic + the alias**

Per connection, bounded by the client's advertised max. Server **MAY**, so it's entirely off unless the client opts in.

## How

- **`write.js`** — `withOutboundTopicAlias` rewrites a v5 PUBLISH just before serialize. It returns a **shallow clone** when aliasing, so the caller's packet — which persistence holds by reference for QoS > 0 and must keep its real topic for resend after reconnect — is never mutated. Non-publish / non-v5 / non-advertising clients hit a cheap early guard (no allocation).
- **`connect.js`** — parse the client's `topicAliasMaximum` into `client._outboundTopicAliasMaximum`; allocate the per-connection alias `Map` only when > 0.
- **Strategy** — assign `1..max` in first-seen order, never evict; once the table is full, later new topics send the full name (no alias). (Eviction/LRU is a possible future optimization.)

## Tests
Register-then-reuse (full topic + alias, then empty topic + same alias); not-advertised (no aliasing); full-table fallback to the full topic.

## Note
The alias map is bounded by `min(clientMax, distinct topics delivered)` — a client advertising `65535` could grow it accordingly (self-inflicted, per connection). A broker-side cap option could be a follow-up if you'd like one.

369 unit tests + lint + tsd green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1129, opened 2026-07-21
feat: MQTT 5.0 support (#1088)

### LINKED ISSUE #783: [question] Return/Reason codes
Hello and thank you for this great broker,

I followed the documentation to implement my broker with authentication, rate limiting, and everything runs smoothly. However, When I close the connection to a client due to rate limits, they have no idea why their connection have cut. As far as they are concerned, it could be their internet connection, or my infrastructure could be down.

I think Reason Codes (MQTT 5) or Return Codes (MQTT 3) are made for that. I've read the MQTT 5 tickets that's mentioning features like that, but not with the current version. Is it possible to do something with the current version ?

You mention that the `packet` can be modified in the `preConnect` function. Is that a lead ? I tried everything, also with the callback, but it never reaches the client. Am I obliged to send a MQTT message to the client before cutting the connection ? That seems a lot of overhead to me.

Thank you for reading, I hope you'll have a idea for me :)

### LINKED ISSUE #822: Reason code on all ACKs

### LINKED ISSUE #823: Reason string on all ACKs

### LINKED ISSUE #824: Message expiry: the lifetime of the Will Message in seconds

### LINKED ISSUE #825: Server disconnect

### LINKED ISSUE #826: Payload format and content type (this should be already handled by mqtt-packet we just have to forward it)

### LINKED ISSUE #827: Request / Response

### LINKED ISSUE #830: User properties (this should be already handled by mqtt-packet we just have to forward it)

### LINKED ISSUE #831: Maximum Packet Size

### LINKED ISSUE #832: Optional Server feature availability

### LINKED ISSUE #834: Subscription options: I think this should already be implemented, expecially the noLocal flag but we should check this

### LINKED ISSUE #836: Server Keep Alive

### LINKED ISSUE #837: Assigned ClientID

### LINKED ISSUE #838: Server reference

### LINKED ISSUE #1090: MQTT 5.0: a PUBLISH matching multiple subscriptions should carry all Subscription Identifiers [MQTT-3.3.4-4]
Deferred follow-up from the MQTT 5.0 review on #1088 (sub-task of #821; related to #828).

### Limitation
[MQTT-3.3.4-4] When a single client has multiple overlapping subscriptions that all match one PUBLISH, the delivered message must carry **all** the matching Subscription Identifiers. aedes delivers a single de-duplicated copy to the client, so only **one** identifier is echoed (the first matching subscription's).

### Why deferred
The fix reaches into the dedupe/delivery core — real regression surface on the hot delivery path, low real-world hit rate. Agreed on #1088 to defer with a documented limitation rather than touch the delivery core in the v5 PR.

### Current state
- Documented in `lib/handlers/subscribe.js` next to the SI-echo logic.
- Single-identifier echo works correctly for the common case (one subscription matches, or one SUBSCRIBE-level identifier).

### Fix sketch
Collect all matching subscriptions' identifiers for a given client+message before/at the dedupe step and emit them as repeated `subscriptionIdentifier` properties on the single delivered PUBLISH.

### LINKED ISSUE #1091: Request / Response Information (CONNECT 0x19 → CONNACK 0x1A)

### LINKED ISSUE #1092: Request Problem Information (CONNECT 0x17)

### LINKED ISSUE #1095: MQTT 3.1.1: zero-length ClientId with CleanSession=0 must be rejected (return code 0x02)
## Summary

For MQTT 3.1.1 (and 3.1.0), a CONNECT with a **zero-length Client Identifier** and **`CleanSession = 0`** must be rejected with CONNACK return code `0x02` (Identifier rejected) and the connection closed. aedes instead **accepts** it and assigns a generated id, regardless of the clean flag.

Surfaced by the new MQTT compatibility workflow (Eclipse Paho interop suite) — `test_zero_length_clientid` fails for v3.1.1.

## Spec

> **[MQTT-3.1.3-7]** If the Client supplies a zero-byte ClientId, the Client MUST also set CleanSession to 1.
>
> **[MQTT-3.1.3-8]** If the Client supplies a zero-byte ClientId with CleanSession set to 0, the Server MUST respond to the CONNECT Packet with a CONNACK return code 0x02 (Identifier rejected) and then close the Network Connection.

(MQTT 3.1.1, §3.1.3.1 Client Identifier.)

## Current behaviour

`lib/handlers/connect.js` assigns an id for any empty client id with no check on the clean flag:

```js
// lib/handlers/connect.js
client.id = clientId || 'aedes_' + randomUUID()
```

None of the preceding `returnCode` checks in `init()` cover the zero-length-id + `clean === false` case, so the connection is accepted.

## Expected

- v3/v4: `clientId.length === 0 && packet.clean === false` → reject with return code `2` (Identifier rejected), then close. (`packet.clean === false` + non-empty id is unaffected; empty id + `clean === true` continues to work and gets a generated id.)
- **v5 is intentionally different and must NOT change:** a zero-length Client Identifier is allowed; the server assigns one and reports it via the Assigned Client Identifier property (already implemented via `client._assignedClientId`). So the new check must be gated on `protocolVersion < 5`.

## Proposed fix sketch

In `init()` (lib/handlers/connect.js), alongside the other `returnCode` checks (before the reject block):

```js
// [MQTT-3.1.3-7/8] v3/v4: a zero-byte ClientId requires CleanSession = 1.
if (packet.protocolVersion < 5 && clientId.length === 0 && packet.clean === false) {
  returnCode = 2 // identifier rejected
}
```

## Reproduction

Eclipse Paho interop suite (`paho.mqtt.testing`), `client_test.py`:

```
python3 client_test.py Test.test_zero_length_clientid --hostname localhost --port 1883
```

Against aedes the first assertion fails (the `cleansession=False` connect is accepted instead of rejected). The same suite is wired up in the MQTT compatibility workflow (#1094).

robertsLando:
## What

Umbrella PR for the `mqttv5` branch → `main`.

MQTT 5.0 support for the broker. MQTT 3.1/3.1.1 behaviour is unchanged throughout.
Feature work does **not** land here directly — each feature is its own reviewable PR
**based on `mqttv5`**. This PR is the single place that tracks overall status and the
issues that get closed when `mqttv5` merges.

- Tracking issue: #821 — per-feature conformance table vs. the
  [OASIS MQTT 5.0 spec](https://docs.oasis-open.org/mqtt/mqtt/v5.0/mqtt-v5.0.html),
  plus the persistence / cluster classification of each item.
- The original umbrella issue #194 is already closed.
- Published for testing as `aedes@2.0.0-beta.1`.

Because the feature PRs target `mqttv5` rather than the default branch, none of the
per-feature issues auto-close as they land — they close here, on merge.

## Issues closed on merge

Everything below is implemented and on the branch today.

- Closes #822 — Reason code on ACKs: CONNACK / SUBACK / UNSUBACK / server-DISCONNECT,
  PUBACK+PUBREC `0x87`, PUBREL/PUBCOMP Success, PUBCOMP `0x92` for an unknown packet id,
  rejecting PUBREC ends the QoS 2 flow without a PUBREL (§4.3.3)
- Closes #823 — Reason String on ACKs: PUBACK / PUBREC (handler `error.message`),
  SUBACK / UNSUBACK / PUBCOMP (broker-generated), gated by Request Problem Information
- Closes #824 — Message Expiry Interval (queued + retained recompute/drop, will-message expiry)
- Closes #825 — Server-initiated DISCONNECT (`0x8E` takeover / `0x8B` shutdown /
  `0x95` packet-too-large / `0x94` topic-alias / `0x82` protocol error), v5-only;
  inbound DISCONNECT `0x04` publishes the will, `0x00` discards it
- Closes #826 — Payload Format Indicator + Content Type (forwarded)
- Closes #827 — Request / Response: Response Topic + Correlation Data forwarded,
  wildcard Response Topic rejected [MQTT-3.3.2-14]
- Closes #830 — User Properties (forwarded)
- Closes #831 — Maximum Packet Size (advertised in CONNACK; oversized inbound rejected
  pre-buffer with `0x95`)
- Closes #832 — Optional server feature availability (Max QoS / Retain / Wildcard /
  Subscription-ID / Shared-Subscription flags)
- Closes #834 — Subscription options: Maximum QoS, No Local, Retain As Published,
  Retain Handling 0/1/2 [MQTT-3.8.3-2]
- Closes #836 — Server Keep Alive
- Closes #837 — Assigned Client Identifier
- Closes #838 — Server Reference / connect-time redirect: `preConnect`/`authenticate`
  reject with `serverReference` → CONNACK `0x9C` / `0x9D` (§4.11)
- Closes #783 — a rejecting `preConnect`/`authenticate` can now carry `reasonCode` +
  `reasonString` to the client instead of a silent socket close
- Closes #1091 — Request / Response Information (CONNECT `0x19` → CONNACK `0x1A`),
  behind the new `responseInformation` broker option
- Closes #1092 — Request Problem Information (CONNECT `0x17`) [MQTT-3.1.2-29]
- Closes #1095 — MQTT 3.1.1 zero-length ClientId with `cleanSession=0` rejected (`0x02`)

### Closes once the in-flight PRs land

Not yet on the branch, so no `Closes` keyword until they merge — this section moves up
when they do:

- #840 Topic Alias — inbound is done; outbound (broker-assigned) is in #1117.
  Merging #1117 completes the issue (follow-up #1124 tracks a broker-side allocation cap).

## Staying open after merge

**No-persistence work still to do** (broker-core, cluster-safe — the batch agreed in
[this comment](https://github.com/moscajs/aedes/issues/821#issuecomment-4893435073)):

- #828 Subscription Identifier — validated, persisted and echoed for a single match, and
  retained-on-subscribe echoes the SI; a delivery matching **multiple** overlapping
  subscriptions still carries only one SI [MQTT-3.3.4-4]. Remaining gap tracked in #1090.
  Same root cause as #1074 (`dedupe()` in `lib/client.js` keeps the first matching
  subscription only, so both its SI *and* its QoS win): the multi-match delivery fix
  closes #1090, #828 and #1074 together.
- #829 Flow control — Receive Maximum parsed, `0` rejected, advertised; **outbound
  in-flight is not capped** (§4.9). Last of the batch: it is the only item on the shared
  delivery hot path, so it lands with a benchmark.
- #833 Enhanced authentication — a CONNECT carrying an Authentication Method is rejected
  with `0x8C` today. #1119 adds the full initial `AUTH` exchange via a new
  `authenticateEnhanced` hook; **re-authentication** (client-initiated `AUTH 0x19`) is a
  fast-follow, so the issue stays open until that lands.

**Persistence / cluster track** (multi-repo — `aedes-persistence` + mongodb/redis/level,
and cluster coordination via `mqemitter-redis` / `-mongodb`); to be picked up as one epic
after the batch above:

- #839 Session Expiry — correct within a running broker (incl. the CONNACK echo of the
  clamped value and inbound-DISCONNECT expiry validation [MQTT-3.14.4-3]); **not durable
  across a broker restart** — needs a persisted deadline + a restart-time reaper
- #835 Will Delay — will persisted, reconnect cancels; the delay timer is per-instance,
  so same durability gap as #839
- #841 Shared Subscriptions — `$share` SUBSCRIBE refused with `0x9E` and
  `sharedSubscriptionAvailable: false` advertised [MQTT-3.2.2-23]. A single-instance
  implementation exists on the `sharedsubs` branch (persistence-backed, **not
  cluster-safe**); a cluster-correct version needs global group membership + deterministic
  recipient selection.
- #821 — the tracking issue itself stays open until the two lists above are empty.

**Known bugs / conformance follow-ups** (none block this merge):

- #1098 — the `willDelayInterval` Will property leaks into the delivered PUBLISH; strict
  v5 clients reject it. Surfaced by the Paho workflow (#1094).
- #1090 — multiple matching Subscription Identifiers (the #828 gap above)
- #1074 — a PUBLISH matching multiple subscriptions should be delivered at the highest
  matching QoS [MQTT-3.3.5-1]. Pre-dates MQTT 5.0 and is untouched by this branch, but it
  shares the #1090 delivery-path root cause, so it is listed here to be fixed alongside it.
- #1096 — test coverage gap: §4.12 robustness + ReasonString / CONNACK capability-flag
  round-trips
- #1124 — broker-side cap for outbound Topic Alias allocation (follow-up to #1117)
- #1125 — honor Maximum Packet Size before appending Reason String / User Properties on ACKs
- Minor, documented: Message Expiry is not decremented on the live delivery path
  (waiting time ≈ 0); the broker→client Maximum Packet Size split is advisory;
  PUBACK/PUBREC emit `0x87` only, not the full set of granular failure codes.

## Landed on `mqttv5`

| PR | What |
|---|---|
| #1088 | MQTT 5.0 core — handshake, property pass-through, inbound Topic Alias, Subscription Identifiers, Session Expiry, server DISCONNECT, Will Delay, Message Expiry, subscription options incl. Retain Handling, Maximum Packet Size, Receive Maximum advertisement, Server Keep Alive, Assigned Client Identifier, `$share` `0x9E`, DoS guards, types + docs |
| #1094 | MQTT 5.0 + 3.1.1 compatibility workflow (pinned Eclipse Paho interop, `tools/mqtt-compat/`) |
| #1097 | MQTT 3.1/3.1.1: reject zero-length ClientId with `cleanSession=0` (#1095) |
| #1100 | Granular ACK reason codes (#822), Reason Strings on PUBACK/PUBREC (#823), Request Problem Information gating (#1092) |
| #1102 | Request / Response Information (#1091) |
| #1115 | Server Reference / connect-time redirect (#838) |
| #1116 | Reason String on the remaining ACKs (#823) + PUBCOMP `0x92` (#822) |
| #1118 | CI: post the MQTT compat comment on fork PRs via `workflow_run` |

## In flight (open, based on `mqttv5`)

| PR | Issue | Notes |
|---|---|---|
| #1117 | #840 | Outbound (broker-assigned) Topic Alias — flips the `test_server_topic_alias` Paho gap |
| #1119 | #833 | Enhanced Authentication (§4.12) `AUTH` exchange via `authenticateEnhanced`; re-auth deferred |

Still to be opened: #828 multi-match Subscription Identifiers, #829 outbound Receive
Maximum enforcement, #833 re-authentication.

## Dependencies — resolved

Both companion packages are published and `package.json` points at published ranges; the
`overrides` shim is gone and `scripts/check-no-git-deps.js` (wired into release-it) passes:

- `aedes-packet@^4.0.1` — moscajs/aedes-packet#192, preserves `properties` + `messageExpiry`
- `aedes-persistence@^10.3.1` — moscajs/aedes-persistence#108, persists `subscriptionIdentifier`

## Tests

- `test/mqtt5.js` — 71 tests driving a real `mqtt@5` client over TCP (plus raw-socket
  tests for the packet-level paths). Full suite green: lint + tsd + unit.
- `tools/mqtt-compat/` — pinned Eclipse Paho interoperability suite on every PR.
  MQTT 3.1.1 is at 100%; MQTT 5.0 climbs as the remaining items land. Current
  `EXPECTED_GAPS`: `test_shared_subscriptions` (#841), `test_flow_control1` (#829),
  `test_server_topic_alias` (#840, in #1117), `test_subscribe_identifiers` (#828).


robertsLando:
@BenjaminDobler here we can keep track of all issues that will be closed when merging this into main

PR #1155, opened 2026-09-16
fix: fix security issue in aedes.js

anupamme:
## Summary
Fix critical severity security issue in `aedes.js`.

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | V-001 |
| **Severity** | CRITICAL |
| **Scanner** | multi_agent_ai |
| **Rule** | `V-001` |
| **File** | `aedes.js:339` |
| **Assessment** | Likely exploitable |
| **Chain Complexity** | 3-step |

**Description**: The Aedes MQTT broker uses a default authentication handler that unconditionally returns success for all connection attempts. This allows any client to connect without providing valid credentials, enabling unauthorized device impersonation and complete broker access.

## Evidence

**Exploitation scenario**: An attacker can connect to the MQTT broker on port 1883 without any authentication.

**Scanner confirmation**: multi_agent_ai rule `V-001` flagged this pattern.

**Production code**: This file is in the production codebase, not test-only code.

## Threat Model Context

This is a Node.js library - vulnerabilities affect downstream consumers who use this package.

## Changes
- `aedes.js`

## Behavior Preservation
The change is scoped to 1 file on the vulnerable path.

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*


robertsLando:
I wouldn't treat this as a real security issue, aedes may be used in test envs where this is not needed

anupamme:
That’s fair. I agree that the default handler is not necessarily a security vulnerability in itself, particularly for test/development deployments where authentication isn’t required.

I intended to flag the potentially surprising behaviour for applications that use Aedes in a production context without realising that the default authenticate handler accepts every client.

The proposed change doesn’t alter that behaviour; it only emits a warning when the default handler is being used. So I think the more accurate framing would be a configuration-safety/developer-warning improvement rather than a critical security fix.

robertsLando:
It should be configurable BTW as some test env could rely on stdout produced by aedes and that line could cause problems or simply devs don't want it repeated every time they start aedes (could happen multiple times). I don't see any benefit sincerly on adding it, I expect devs using aedes know that if they don't setup any auth rule there will be no auth rule

